Issue #14, opened 2018-10-05
Add support for BankID through GrandID (Svensk E-Identitet)


PeterOrneholm:
We aim to support BankID not only natively but also through GrandID (Svensk E-Identitet) as that has advantages such as being quicker on getting started with.

The GrandID implementation will follow the same principles as for the BankID, such as that the API wrapper will be a separate project and NuGet package,

PeterOrneholm:
Implemented in #15.

PR #15, opened 2018-10-05
Implement GrandID as described in #14

PeterOrneholm:
This PR implements support for GrandID (Svensk E-Identitet) as that has advantages such as being quicker on getting started with when using BankID.

The GrandID implementation follows the same principles as for the BankID, such as that the API wrapper is a separate project and NuGet package,

A simple usage of it looks like this:

```csharp
services.AddAuthentication()
    .AddGrandId(builder =>
    {
        builder
            .UseProdEnvironment(Configuration.GetValue<string>("ActiveLogin:GrandId:ApiKey"))
            .AddScheme("grandid-choosedevice", "GrandID - ChooseDevice", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
            });
    });
```

A more advanced configuration could look like this:

```csharp
services.AddAuthentication()
    .AddGrandId(builder =>
    {
        if (Configuration.GetValue("ActiveLogin:GrandId:UseDevelopmentApi", false))
        {
            builder.UseDevelopmentEnvironment("Alice", "Smith");
        }
        else
        {
            var apiKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ApiKey");
            if (Configuration.GetValue("ActiveLogin:GrandId:UseTestApiEndpoint", false))
            {
                builder.UseTestEnvironment(apiKey);
            }
            else
            {
                builder.UseProdEnvironment(apiKey);
            }
        }

        builder
            .AddScheme("grandid-samedevice", "GrandID - SameDevice", options =>
            {
                options.CallbackPath = new PathString("/signin-grandid-samedevice");
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
            })
            .AddScheme("grandid-otherdevice", "GrandID - OtherDevice", options =>
            {
                options.CallbackPath = new PathString("/signin-grandid-otherdevice");
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:OtherDeviceServiceKey");
            })
            .AddScheme("grandid-choosedevice", "GrandID - ChooseDevice", options =>
            {
                options.CallbackPath = new PathString("/signin-grandid-choosedevice");
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
            });
    });
```

And if you only want to try it out in memory for demo purposes, you could use this:

```csharp
services.AddAuthentication().AddGrandId();
```


src/ActiveLogin.Authentication.GrandId.AspNetCore/GrandIdAuthenticationBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace ActiveLogin.Authentication.GrandId.AspNetCore
+{
+    public class GrandIdAuthenticationBuilder
+    {
+        public AuthenticationBuilder AuthenticationBuilder { get; }
+
+        private readonly List<Action<HttpClient>> _httpClientConfigurators = new List<Action<HttpClient>>();
+        private readonly List<Action<HttpClientHandler>> _httpClientHandlerConfigurators = new List<Action<HttpClientHandler>>();
+
+        public IServiceCollection Services;
```
fredrik-lundin: Not sure that `Services` should be exposed publicly at this level, it can be accessed via `AuthenticationBuilder.Services` anyway. Accessing via `AuthenticationBuilder` also gives a better hint that we are not in any way mixing service collections. 

If we for some reason want to keep it public, it should at least be `get` only (just like we do it for `AuthenticationBuilder`)   in my opinion.

PeterOrneholm: I agree. Fixed now.

fredrik-lundin:
I like how `services.AddAuthentication().AddGrandId()` spins up the development environment out of the box. But a tiny concern with it is that this extension would be inconsistent with the `AddBankId()` extension where one has to add an extra service call, like this:

```
services.AddAuthentication()
        .AddBankId();

services.AddBankIdDevelopmentEnvironment();
```

Should one of them be changed to have a more consistent api for users of the library, or is this not a problem?

fredrik-lundin:
A personal opinion about the api design is that I think we should provide better defaults instead of making it super flexible with the `AddScheme` methods. 

Since there are not an unknown number of schemes they can provide for grandid (I guess?), we can provide help methods for the three different cases instead of having the user specify everything themselves with `AddScheme`. Something like this:

```cs
services.AddAuthentication()
                .AddGrandId(builder =>
                {
                    // ...

                    builder
                        .AddSameDevice("my-custom-scheme", "My custom display name", options =>
                        {
                            options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
                        })
                        .AddOtherDevice(options =>
                        {
                            options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:OtherDeviceServiceKey");
                        })
                        .AddChoosenDevice(options =>
                        {
                            options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
                        });
                });
``` 

If we do it like that, we can even let the user leave the `options.AuthenticationServiceKey` blank, and if it's blank, we can look for the correct value in settings by convention (as an alternative to specifying it explicitly in `Startup` of course, the option to put it in there "manually" should not go away).

I think this makes it easier to use the api, hides a lot of defaults that the user doesn't really HAVE to care about (like `CallbackPath`), while still leaving it as configurable as before, if someone wants that freedom. 

See for example how the `AddTwitter` extension is implemented, with defaults in mind: https://github.com/aspnet/Security/blob/master/src/Microsoft.AspNetCore.Authentication.Twitter/TwitterExtensions.cs

PeterOrneholm:
> Since there are not an unknown number of schemes they can provide for grandid (I guess?), we can provide help methods for the three different cases instead of having the user specify everything themselves with `AddScheme`.

I like! In theory, GrandID supports more scenarios than just BankID, like Siths and username/password. But we would not support whatever response they have (as they don't return a personal identity number for example). Therefore it's those three available at the moment.

I'd suggest having `.AddScheme()` as a base and then add three extension methods for `.AddSameDevice()`, `.AddOtherDevice()` and `.AddChooseDevice()`.

I'd suggest these variants, and in the case where you leave the keys out, it will look at a predefined setting. The same could then go for the API key.

```
.AddSameDevice("my-custom-scheme", "My custom display name", options =>
{
 options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
})
.AddSameDevice("My custom display name", options =>
{
 options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
})
.AddSameDevice("My custom display name", Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey"))
.AddSameDevice("My custom display name")
```

What do you think?

fredrik-lundin:
> In theory, GrandID supports more scenarios than just BankID, like Siths and username/password. 

Very good point, I forgot about that. But having typed schemes would make it even easier in the long run to add something like `AddSiths()`, since that might need some special treatments when it comes to parsing the result from grand id etc.

> I'd suggest having `.AddScheme()` as a base and then add three extension methods for `.AddSameDevice()`, `.AddOtherDevice()` and `.AddChooseDevice()`.
> 
> I'd suggest these variants, and in the case where you leave the keys out, it will look at a predefined setting. The same could then go for the API key.
> 
> What do you think?

I think `.AddScheme()` should stay only if there are any realistic use-cases for anyone to use it. I can't think of any from the top of my head, but, or someone else might. If you do, then it should definitely stay. But if we can't think of any such scenario, if think it should be hidden som the consumer of the library as it will only add more confusion to how to configure this the correct way.

fredrik-lundin:
Sounds good, great job! :)

fredrik-lundin:
After doing some pair programming with @PeterOrneholm, we landed with the following api:

```csharp
services.AddAuthentication()
    .AddGrandId(builder =>
    {
        // set environment etc ...

        // Set options (with AuthenticateServiceKey)
        builder
            .AddSameDevice(options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
            })
            .AddOtherDevice(options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:OtherDeviceServiceKey");
            })
            .AddChooseDevice(options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
            });

        //// Set AuthenticationScheme and options (with AuthenticateServiceKey)
        builder
            .AddSameDevice("grandid-samedevice-custom", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
            })
            .AddOtherDevice("grandid-otherdevice-custom", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:OtherDeviceServiceKey");
            })
            .AddChooseDevice("grandid-choosedevice-custom", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
            });

        //// Set AuthenticationScheme, DisplayName and options (with AuthenticateServiceKey)
        builder
            .AddSameDevice(GrandIdAuthenticationDefaults.SameDeviceAuthenticationScheme, "GrandID - SameDevice", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:SameDeviceServiceKey");
            })
            .AddOtherDevice(GrandIdAuthenticationDefaults.OtherDeviceAuthenticationScheme, "GrandID - OtherDevice", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:OtherDeviceServiceKey");
            })
            .AddChooseDevice(GrandIdAuthenticationDefaults.ChooseDeviceAuthenticationScheme, "GrandID - ChooseDevice", options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
            });
    });

// Sample of using BankID through GrandID (Svensk E-identitet) with minimum configuration
services.AddAuthentication()
    .AddGrandId(builder =>
    {
        builder
            .UseProdEnvironment(Configuration.GetValue<string>("ActiveLogin:GrandId:ApiKey"))
            .AddChooseDevice(options =>
            {
                options.AuthenticateServiceKey = Configuration.GetValue<string>("ActiveLogin:GrandId:ChooseDeviceServiceKey");
            });
    });
```

The reason for this was:
1. It follows standard conventions for adding external authentication providers
2. It exposes valid use-cases of the api in a more explicit way. Making it easier for the end user to use the library in a correct way
3. It will make it easier to add more Grand ID functionality in the future, such as SITHS authentication

Issue #16, opened 2018-10-09
Refactor API design for BankID to harmonize with GrandID


PeterOrneholm:
During the development of the GrandID authentication handler a thorough work went into making sure the API, especially the public API, being developer friendly as well ass following common conventions.
 
We aim to refactor the BankID authentication handler before reaching stable 1.0.0 to use the same API design.

This issue will keep track of that work and, before implementing it, will have a suggested API design described here.

PeterOrneholm:
The proposed API design looks like this:

```csharp
services.AddAuthentication()
        .AddBankId(builder => {
        builder
            .UseProdEnvironment()
            .UseBankIdClientCertificateFromAzureKeyVault(options =>
            {
                options.AzureAdClientId = Configuration.GetValue<string>("ActiveLogin:BankId:ClientCertificate:AzureAdClientId");
                options.AzureAdClientSecret = Configuration.GetValue<string>("ActiveLogin:BankId:ClientCertificate:AzureAdClientSecret");
                options.AzureKeyVaultSecretIdentifier = Configuration.GetValue<string>("ActiveLogin:BankId:ClientCertificate:AzureKeyVaultSecretIdentifier");
            })
            .UseBankIdRootCaCertificate(Path.Combine(_environment.ContentRootPath, Configuration.GetValue<string>("ActiveLogin:BankId:CaCertificate:FilePath")))
            .Add("CustomAuthScheme", "CustomDisplayName", options =>
            {
                // ...
            });
        });
```

With shorthand available for the client certificate options:

```csharp
// ...
.UseBankIdClientCertificateFromAzureKeyVault(Configuration.GetSection("ActiveLogin:BankId:ClientCertificate"))
// ...
```

My biggest concern is what to call the `.Add()` method. In GrandId they support three distinct scenarios (`.AddSameDevice()`, `.AddOtherDevice()`, `.AddChooseDevice()`) and native BankID does not work exactly like that. In the future we might be able to achieve the same using policies and device detection, but we don't at the moment.

Therefore we need something that just adds the BankID scheme with options. Possible names as I see it:
* `.Add()`
* `.AddInstance()`
* `.AddScheme()`

Please provide input and vote on a possible name.

fredrik-lundin:
Are there any obvious use-cases where one would like to add more than one scheme?

PeterOrneholm:
Adding some background to why we need / want to redesign:

[...]

PR #23, opened 2018-10-14
Redesign BankID auth design

PeterOrneholm:
This PR redesigns BankID API design to match the one in GrandID as described in #16.
To be able to achieve same API design some missing features are implemented in this PR, such as AutoLaunch and the possibility to hide the input for PIN.

It also solves #26 and uses the small set of code needed as private assets instead of public NuGet package,


src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationBuilderEnvironmentExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public static IBankIdAuthenticationBuilder UseDevelopmentEnvironment(this IBankIdAuthenticationBuilder builder, string givenName, string surname)
+            => UseDevelopmentEnvironment(builder, x => new BankIdDevelopmentApiClient(givenName, surname));
+
+        public static IBankIdAuthenticationBuilder UseDevelopmentEnvironment(this IBankIdAuthenticationBuilder builder, string givenName, string surname, string personalIdentityNumber)
+            => UseDevelopmentEnvironment(builder, x => new BankIdDevelopmentApiClient(givenName, surname, personalIdentityNumber));
+
+        private static IBankIdAuthenticationBuilder UseDevelopmentEnvironment(this IBankIdAuthenticationBuilder builder, Func<IServiceProvider, IBankIdApiClient> bankIdDevelopmentApiClient)
+        {
+            builder.AuthenticationBuilder.Services.AddSingleton(bankIdDevelopmentApiClient);
```
nikolaykrondev: Lets follow the same pattern as we have already with TryAdd
```suggestion
            builder.AuthenticationBuilder.Services.TryAddSingleton(bankIdDevelopmentApiClient);
```

src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationBuilderSchemeExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+            => AddCustom(builder, authenticationScheme, BankIdAuthenticationDefaults.CustomDisplayName, configureOptions);
+
+        public static IBankIdAuthenticationBuilder AddCustom(this IBankIdAuthenticationBuilder builder, string authenticationScheme, string displayName, Action<BankIdAuthenticationOptions> configureOptions)
+            => AddScheme(builder, authenticationScheme, displayName, configureOptions, options =>
+            {
+                options.CallbackPath = BankIdAuthenticationDefaults.CustomCallpackPath;
+            });
+
+
+        public static IBankIdAuthenticationBuilder AddSameDevice(this IBankIdAuthenticationBuilder builder, Action<BankIdAuthenticationOptions> configureOptions)
```
nikolaykrondev: Could we may be add one more overload method for BankId that does not take any arguments, or make configuredOptions equal null for the selected one. Then we can simplify the template code to:
```
.AddBankId(builder =>
                {
                    builder
                        .UseDevelopmentEnvironment()
                        .AddSameDevice()
                        .AddOtherDevice();
                });
```

samples/IdentityServerSample/Controllers/AccountController.cs
```diff
@@ -67,7 +67,7 @@ public async Task<IActionResult> ExternalLoginCallback()
 
             if (_interaction.IsValidReturnUrl(returnUrl) || Url.IsLocalUrl(returnUrl))
             {
-                return Redirect(returnUrl);
+                return LocalRedirect(returnUrl);
```
nikolaykrondev: Might be a mistake, we do not take all the time returnUrl as a local one, first value in the condition does not guaranty that
```suggestion
                return Redirect(returnUrl);
```

PR #30, opened 2018-10-25
Add support for setting a personal identity number from code

nikolaykrondev:


src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationHandler.cs
```diff
@@ -139,7 +139,10 @@ protected override Task HandleChallengeAsync(AuthenticationProperties properties
 
             var loginOptions = new BankIdLoginOptions(
                 Options.BankIdCertificatePolicies,
-                null, 
+                properties.Items.TryGetValue(
```
PeterOrneholm: Can this be refactored into a method so it's clearer what it does? The onliner is a bit hard to get a grasp of at the moment.

src/ActiveLogin.Authentication.GrandId.AspNetCore/GrandIdAuthenticationHandler.cs
```diff
@@ -142,9 +142,12 @@ protected override async Task HandleChallengeAsync(AuthenticationProperties prop
             AppendStateCookie(properties);
 
             var absoluteReturnUrl = GetAbsoluteUrl(Options.CallbackPath);
+            properties.Items.TryGetValue(
```
PeterOrneholm: I'd like to refactor this as it's own method as well, for clarity.

src/ActiveLogin.Authentication.GrandId.Api/GrandIdDevelopmentApiClient.cs
```diff
@@ -43,6 +43,11 @@ public async Task<FederatedLoginResponse> FederatedLoginAsync(FederatedLoginRequ
         {
             await SimulateResponseDelay().ConfigureAwait(false);
 
+            if (!string.IsNullOrEmpty(request.PersonalIdentityNumber))
```
PeterOrneholm: This will cause concurrency issues as the dev implementation is registered as singleton so the _personalIdentityNumber would end up changing if someone logs in. Needs to stay readonly.

PeterOrneholm:
I know @fredrik-lundin will be auser of this feature, so would like to get his feedback as well.

src/ActiveLogin.Authentication.GrandId.Api/GrandIdDevelopmentApiClient.cs
nikolaykrondev: True, completely forgot that this one is singleton. In that case we need to add such parameter to `SessionStateRequest` and populate it from `HandleRemoteAuthenticateAsync`, any other ideas?

PeterOrneholm: You can just add it to the list of federatedRequests. Just need to slightly modify what that contains. I can show later when I have access to a computer.

PeterOrneholm:
@nikolaykrondev Please have a look at my commits. I fixed the dev and a few other things. Also normalized the pnr before sending it to GrandID as they require the long version.

nikolaykrondev:
Yeah, I see.
Nice, good job!
Should we merge?

fredrik-lundin:
LGTM

PR #31, opened 2018-10-25
Normalize the names returned from BankID and GrandID

nikolaykrondev:
I did not put the shared method to Common since this is a PoC with ToTitleCase with only one row, otherwise it creates to much over code to instantiate the helper based on Options.

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationHandler.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationHandler.cs
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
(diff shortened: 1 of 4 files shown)


PeterOrneholm:
Following up on the discussion in https://github.com/ActiveLogin/ActiveLogin.Authentication/issues/20-
I'm not really sure if we should actually implement this. From your testing you said there are quite a few scenarios not handled right with the implementation we found. I'd say it might actually cause more issues than it solves.

If we were to merge it we'd need to add test to show what we support and what we don't.

But, if not at least 99% of names works, we should maybe skip it?

nikolaykrondev:
I agree... would be nice to have a poll or some another way to ask community what would they wish

PeterOrneholm:
I’d vote for writing a small doc under FAQ instead. So if someone wants to normalize we can show that code is a good start. Then we populate the claims with raw values but a consumer can normalize later.

PeterOrneholm:
We close this one and use the documentation implemented in #37 instead.

PR #32, opened 2018-10-25
Add a target for net461

nikolaykrondev:


nikolaykrondev:
We need to update all the references to Identity package that supports 461 before merging this

PeterOrneholm:
I've pushed new aplha version to NuGet that targets net461

PR #43, opened 2018-11-29
Remove hardcoded auth schemas

nikolaykrondev:
Reviewed the schemas that we use and tried to avoid hardcoding. Using predefined defaults now


PeterOrneholm:
Great fix @nikolaykrondev !

PR #44, opened 2018-12-03
Siths support

hematmedhelp:
Lesser used "device" within GrandId.

Library threw exception when trying to parse and empty social security number - which is not an included claim when authenticating with Siths.

Also added the userAttributes from this flow/response.


hematmedhelp:
UseDevelopment didn't pan out well. Will fix.

PeterOrneholm:
Hi @hematmedhelp! Fantastic that you are helping out with Active Login and are doing a PR! SITHS support is something we are planning on doing in a future release. Maybe 1.1.0 or later. I've added an issue (#45) where we can discuss how this should be implemented.

I'd like to discuss how it should be implemented before, like if we instead maybe should have .AddGrandIdBankId() and .AddGrandIdSiths() to make a clear distinction. Please update the issue with a proposed solution and we'll have a look :)

Issue #46, opened 2018-12-03
Naming of .AddGrandId()


PeterOrneholm:
Preparing for supporting SITHS through GrandID (#45) inspired by the PR #44 by @hematmedhelp I realized it might be a good thing to rename the .AddGrandId() method into something more specific for BankID.

BankID and SITHS do share some parts in the underlying API (like name etc) but differs in what other data is returned (Personal Identty number, email etc). Also, concepts like .AddSameDevice() and .AddOtherDeivce() probably differs enough that it should be two implementations, that might (and will) share code under the covers.

I see two potential patterns:

**Pattern 1:**

* `.AddGrandIdBankId()`
* `.AddGrandIdSiths()`

**Pattern 2:**

An alternative would be to keep the top level `.AddGrandId()` but split it further down, like this:

```csharp
services
    .AddAuthentication()
    .AddGrandId(builder =>
    {
        builder
            .UseDevelopmentEnvironment()
            .AddBankIdSameDevice(options => { })
            .AddBankIdOtherDevice(options => { })
            .AddSiths(options => { });
    });
```

Any input or other ideas?

PeterOrneholm:
Would be great with input from @hematmedhelp who wrote the PR and from @fredrik-lundin which I know has worked with SITHS before.

Zonnex:
If Siths is something related to GrandId, then I'd expect it to be part of the builder

hematmedhelp:
@PeterOrneholm 

According to the API-documentation from e-identitet the flows and endpoints for GrandId and Siths are identical - the practical difference that I've found is the reply with the user claims.

In fact Siths will work currently in any of the methods within `AddGrandId()` as long as you change this line:

```
 private IEnumerable<Claim> GetClaims(SessionStateResponse loginResult, DateTimeOffset? expiresUtc)
        {
            // will throw exception due to PersonalIdentityNumber being null
            var personalIdentityNumber = SwedishPersonalIdentityNumber.Parse(loginResult.UserAttributes.PersonalIdentityNumber);
           ...
       }
```

So I'm guessing this discussion is more about the syntax rather than the actual implementation.

Since Siths is part of GrandId I think **Pattern 2** is more representable.

hematmedhelp:
But perhaps **Pattern 2** isn't the most viable. Since the options from the builder might become confusing. Things like `IssueGenderClaim` isn't relevant and won't work since you're not getting a social security number to verify against.

[...]

PR #47, opened 2018-12-06
Add BankId prefix to GrandId things that handles BankId

PeterOrneholm:
As discussed in #46 we should rename some things for GrandId to explicitly state that they are used for the BankId part to support Siths in a future release.


PeterOrneholm:
The PR is now updated.
The methods on the API are now splitted so that only those parameters / result relevant to the usecase are present. This will enable Siths to have a more explicit API.

The Handler is splitted into a base and a BankId implementation, making it easy to have a Siths implementation.

src/ActiveLogin.Authentication.GrandId.AspNetCore/GrandIdAuthenticationBuilderExtensions.cs
```diff
@@ -4,7 +4,6 @@ public static class GrandIdAuthenticationBuilderExtensions
     {
         internal static IGrandIdAuthenticationBuilder AddDefaultServices(this IGrandIdAuthenticationBuilder builder)
```
martinsjolanders: AddDefaultServices doesn't seem to do anything, mayby remove?

martinsjolanders:
Should change 
return string.Join("&", queryStringParams.Select(x => $"{x.Key}={Uri.EscapeDataString(x.Value)}"));
to
QueryBuilder queryBuilder = new QueryBuilder(queryStringParams);
return queryBuilder.ToQueryString().ToString();
at
https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/06bf812e6cae79f9f68c614814a4d1d7a18a3a65/src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdLauncher.cs#L59

src/ActiveLogin.Authentication.GrandId.AspNetCore/GrandIdAuthenticationDefaults.cs
```diff
@@ -5,21 +5,22 @@ namespace ActiveLogin.Authentication.GrandId.AspNetCore
...
-        public const string ChooseDeviceDisplayName = "BankID";
-        public static readonly PathString ChooseDeviceCallpackPath = new PathString("/signin-grandid-choosedevice");
+        public const string BankIdOtherDeviceAuthenticationScheme = "grandid-bankid-otherdevice";
+        public const string BankIdOtherDeviceDisplayName = "BankID (Annan enhet)";
+        public static readonly PathString BankIdOtherDeviceCallpackPath = new PathString("/signin-grandid-bankid-otherdevice");
 
-        public const string IdentityProviderName = "GrandId";
-        public const string AuthenticationMethodName = "grandid";
+        public const string BankIdChooseDeviceAuthenticationScheme = "grandid-bankid-choosedevice";
+        public const string BankIdChooseDeviceDisplayName = "BankID";
```
martinsjolanders: Should be BankID (Välj enhet)

PeterOrneholm: Thanks for input!

I've been thinking about naming in that, but my conclusion is that you will probably wither display "BankID (Denna enhet)" + "BankID (Annan enhet)" or only display "BankID (Välj enhet)", no reason to display them all three at once. And therefore, if it's only one BankID option available, it makes more sense to only call it BankID as I see it. What do you think?

PeterOrneholm:
> Should change:

Fixed!

PR #54, opened 2018-12-26
Test BankID API and GrandID API and some refactorings

PeterOrneholm:
As we were lacking quite a lot of tests for the APIs I've now created a suite of them (~
50 in total) testing serialization, mappings, error handling, development client etc.

While writing all of the tests I found some good refactorings that could be made before going 1.0 as they affected the public API.

These include, but are not limited to:

- Make classes and constructors internal when no public access was needed
- Make properties get only when no public set was needed
- Rename some public methods to be consistent
- Move logic out of the models into extension methods to expose the raw values returned from the API and then build parsing and logic on top of that.


test/ActiveLogin.Authentication.BankId.Api.Test/BankIdApiClientExtensions_Tests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace ActiveLogin.Authentication.BankId.Api.Test
+{
+    public class BankIdApiClientExtensions_Tests
+    {
+        [Fact]
+        public async void AuthAsync_WithEndUserIp_ShouldMap_ToAuthRequest_WithEndUserIp()
+        {
+            // Arrange
+            var bankIdApiClientMock = new Mock<IBankIdApiClient>(MockBehavior.Strict);
```
nikolaykrondev: Was the strict mode chosen by purpose? :)

nikolaykrondev: Hard to say, I started reading about it and found an article with the debates, it is like tabs vs spaces :) I think in that case is better to keep it strict - would be much easier to track failed tests if we add new logic in some methods (like method calling another method - this is the most valuable argument in these debates)

src/ActiveLogin.Authentication.BankId.Api/Models/CertExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace ActiveLogin.Authentication.BankId.Api.Models
+{
+    public static class CertExtensions
+    {
+        private static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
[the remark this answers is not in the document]
PeterOrneholm: @nikolaykrondev I pused it now, found 2 places where it could be used. One for "parsing" and one for serializting :)

Issue #84, opened 2019-02-13
Add support for sign in BankId api client


ellinge:
The BankId API client only expose the "auth" method but not the similar "sign" with the extra properties in the request as described in **14.1.2 Additional Parameters for sign**.
I know the package is focused on Authentication, but I think it should expose the sign-functionality as well since it's almost there.

PeterOrneholm:
Great suggestion and thanks for the PR! I've added a few comments.

PeterOrneholm:
Fixed in #85.

PeterOrneholm:
Finally released :)
https://www.nuget.org/packages/ActiveLogin.Authentication.BankId.Api/2.0.0-rc-1

Please give it a try before we go 2.0.0!

ellinge:
Ping @PeterOrneholm 
![bild](https://user-images.githubusercontent.com/17863113/57050505-35c00800-6c7d-11e9-91bd-2ba5a1bb7e39.png)

Seems to be working fine. Logged in as well before signing.

PeterOrneholm:
> Seems to be working fine. Logged in as well before signing.

Nice! Will try a few things before going 2.0.0 sharp. Thanks for testing the RC!

PR #85, opened 2019-02-13
#84 Feature/add sign to bankid api

ellinge:
Have not added complete simulation but I think this should be enough to start of with for #84.


src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
```diff
@@ -94,7 +95,23 @@ public async Task<AuthResponse> AuthAsync(AuthRequest request)
             return new AuthResponse(orderRef, autoStartToken);
         }
 
-        private string GetPersonalIdentityNumber(AuthRequest request)
+        public async Task<SignResponse> SignAsync(SignRequest request)
```
PeterOrneholm: Would like to see a refactoring where the shared logic between AuthAsync and SignAsync are placed in a private method. I think they are identical in what they do, but are working with different input / output.

src/ActiveLogin.Authentication.BankId.Api/Models/IAuthRequest.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace ActiveLogin.Authentication.BankId.Api.Models
+{
+    public interface IAuthRequest
```
PeterOrneholm: I know that Auth and Sign share these properties, but would like to keep the requests as clean as possible, every public interface / class we add needs to be supported.

I don't think that the interface adds any value at the moment, except for the simulated client. I'd vote for removing the interface and handle the scenario manually.

src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
```diff
@@ -94,7 +95,23 @@ public async Task<AuthResponse> AuthAsync(AuthRequest request)
...
+            var orderRef = Guid.NewGuid().ToString();
+            var auth = new Auth(request.EndUserIp, orderRef, personalIdentityNumber);
+            _auths.Add(orderRef, auth);
+
+            var autoStartToken = Guid.NewGuid().ToString().Replace("-", string.Empty);
+
+            return new SignResponse(orderRef, autoStartToken);
+        }
+
+        private string GetPersonalIdentityNumber(IAuthRequest request)
```
PeterOrneholm: Take `string personalIdentityNumber` as input instead as I'm suggesting removing the `IAuthRequest `interface. See comment below.

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
```diff
@@ -0,0 +1,139 @@
...
+        [DataMember(Name = "userVisibleData")]
+        public string UserVisibleData { get; private set; }
+
+        /// <summary>
+        /// Data not displayed to the user. String. The value must be base 64-encoded. 1-200 000 characters after base 64-encoding.
+        /// </summary>
+        [DataMember(Name = "userNonVisibleData", EmitDefaultValue = false)]
+        public string UserNonVisibleData { get; private set; }
+
+        private static string EnsureBase64EncodedString(string value)
```
PeterOrneholm: I would like us to always expect the user to send in the data as a "normal string" and that the Base64 encoding is always done. That would be much more predictable in behavior. So, rename to `ToBase64String` and donät check if already Base64.

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            if (value == null) return null;
```
PeterOrneholm: This isn't documented, but we usually do multi line if statements for this, so:
```suggestion
            if (value == null) {
                return null;
            }
```

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace ActiveLogin.Authentication.BankId.Api.Models
+{
+    /// <summary>
+    /// Sign request parameters.
+    /// </summary>
+    [DataContract]
+    public class SignRequest : IAuthRequest
```
PeterOrneholm: I see an issue with adding/removing parameters in the middile when they are all of the same type (personalIdentity Number).

Suggesting these overloads:

- `public SignRequest(string endUserIp, string userVisibleData)`
- `public SignRequest(string endUserIp, string userVisibleData, string userNonVisibleData)`
- `public SignRequest(string endUserIp, string userVisibleData, string userNonVisibleData, string personalIdentityNumber)`
- `public SignRequest(string endUserIp, string userVisibleData, string userNonVisibleData, string personalIdentityNumber, Requirement requirement)`

test/ActiveLogin.Authentication.BankId.Api.Test/BankIdApiClientExtensions_Tests.cs
```diff
@@ -1,15 +1,17 @@
...
+
 // ReSharper disable InvokeAsExtensionMethod
 
 namespace ActiveLogin.Authentication.BankId.Api.Test
 {
     public class BankIdApiClientExtensions_Tests
     {
         [Fact]
-        public async void AuthAsync_WithEndUserIp_ShouldMap_ToAuthRequest_WithEndUserIp()
+        public async Task AuthAsync_WithEndUserIp_ShouldMap_ToAuthRequest_WithEndUserIp()
```
PeterOrneholm: Thanks, nice catch on the async voids!

src/ActiveLogin.Authentication.BankId.Api/BankIdApiClientExtensions.cs
```diff
@@ -48,6 +48,58 @@ public static Task<AuthResponse> AuthAsync(this IBankIdApiClient apiClient, stri
             return apiClient.AuthAsync(new AuthRequest(endUserIp, personalIdentityNumber));
         }
 
+        /// <summary>
```
PeterOrneholm: See comment below on overloads for construvtor on SignRequest. The same I'd say apply here, but maybe we don't need all overloads?

Suggestion:
- `public SignRequest(string endUserIp, string userVisibleData)`
- `public SignRequest(string endUserIp, string userVisibleData, string userNonVisibleData)`
- `public SignRequest(string endUserIp, string userVisibleData, string userNonVisibleData, string personalIdentityNumber)`

src/ActiveLogin.Authentication.BankId.Api/BankIdApiClientExtensions.cs
```diff
@@ -48,6 +48,58 @@ public static Task<AuthResponse> AuthAsync(this IBankIdApiClient apiClient, stri
...
+        /// <param name="endUserIp">
+        /// The user IP address as seen by RP.String.IPv4 and IPv6 is allowed.
+        /// Note the importance of using the correct IP address.It must be the IP address representing the user agent (the end user device) as seen by the RP.
+        /// If there is a proxy for inbound traffic, special considerations may need to be taken to get the correct address.
+        ///
+        /// In some use cases the IP address is not available, for instance for voice based services.
+        /// In this case, the internal representation of those systems IP address is ok to use.
+        /// </param>
+        /// <param name="userVisibleData">
+        /// The text to be displayed and signed. String. The text can be formatted using CR, LF and CRLF for new lines. The text must be encoded as UTF-8 and then base 64 encoded. 1--40 000 characters after base 64 encoding.
```
PeterOrneholm: See comment below on Base64-handling. Therefore the XML doc should not mention the Base64 part. All XML-docs for `userVisibleData` should be updated to:
`The text to be displayed and signed. The text can be formatted using CR, LF and CRLF for new lines.`

src/ActiveLogin.Authentication.BankId.Api/BankIdApiClientExtensions.cs
```diff
@@ -48,6 +48,58 @@ public static Task<AuthResponse> AuthAsync(this IBankIdApiClient apiClient, stri
...
+        /// </param>
+        /// <param name="userNonVisibleData">
+        /// Data not displayed to the user. String. The value must be base 64-encoded. 1-200 000charactersafter base 64-encoding.
```
PeterOrneholm: See comment below on Base64-handling. Therefore the XML doc should not mention the Base64 part. All XML-docs for `userNonVisibleData` should be updated to:
`Data not displayed to the user.`

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
ellinge: If you want to base64 something other than a string you won' be able to though? In that case I'd suggest you change the input parameter type or add a overload with object userNonVisibleData at least which handles that?

PeterOrneholm: Very good point! Let me have a look on other implementations and get back.

The userVisibleData is always a string in the end (from what I can read), so that is more straight forward, but on the other hand it might be strange if we encode one but not the other.

Do you know what the userNonVisibleData is used for? Is it returened on collect in any way?

ellinge: I don't have a response at hand from the new endpoint, another project today, but this is from my notes when investigating the v4 soap endpoint. Unfortunately, in Swedish.
bild

I hopefully have some time tonight to implement your suggestions and will give you a sample from the new endpoint. I'm also thinking about adding a Raw collect-method as well if you want to store the complete original response as is without deserializing.

ellinge: Maybe it would be easier just to change all input parameters to byte[] instead for the invisible part, Then the user can choose freely. It would simplify the overloads/constructors as well, not everything being string. How do you stand when it comes to optional parameters as well, it would simplify when it comes to the number of explicit constructors needed.

ellinge: Changed to byte[] now instead

ellinge: bild
Visible in bankidprompt: "test signatur" Non visible: new byte[1]

Fetched with CompletionData.GetSignatureXml()

fredrik-lundin:
Hi,

Looks like you've done a massive job on this PR. Thanks so much for that! 

I haven't had time to get into the code yet unfortunately. Can you give us a short description on how the usage of this code is gonna look? 

Any specifik setup during startup? And how is the actual signing calls looking?

ellinge:
Even if it looks massive it's quite simple. Just added support for calls to /sign. The workflow is identical to /auth except the following:
* two additional parameters. One visible to the user during the signing phase and one optional non visible. The guideline recommendation is to use a hash when signing a file to accomodate all file sizes. But could be an text extract to the complete model as json during signing or anything covering your needs. 
* when the signing process is done collect the signing result with CompletionData.GetSignatureXml()
It contains everything you need to prove the signing took place with the parameters used and by whom.

I only use the client-part. Something which would be nice is some of the HttpClient setup parts where moved to the client-library instead. Now I've manually initialized the HttpClient. 

We store our certificates in the windows certificate store so that's a nice alternative to the disk based, if you don't use azure. But that's a separate PR/issue.

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// In this case, the internal representation of those systems IP address is ok to use.
+        /// </param>
+        /// <param name="userVisibleData">
+        /// The text to be displayed and signed. The text can be formatted using CR, LF and CRLF for new lines.
+        /// </param>
+        /// <param name="personalIdentityNumber">
+        /// The personal number of the user. 12 digits, century must be included (YYYYMMDDSSSC).
+        /// If the personal number is excluded, the client must be started with the AutoStartToken returned in the response.
+        /// </param>
+        public SignRequest(string endUserIp, string userVisibleData, string personalIdentityNumber)
```
PeterOrneholm: Great idea with the byte[] on userNonVisibleData.
I'd suggest removing this overload, as it would hinder us to do an overload that take userNonVisibleData as string if we see a need for that in the future.

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
PeterOrneholm: Thanks for the samples!

PeterOrneholm:
> First off. Thanks for the repo to start off with. 

Thank you, I like how we can collaborate on things like these :) Out of interest, can you disclose what project you are using it for?

> I only use the client-part. Something which would be nice is if some of the HttpClient setup parts where moved to the client-libraries instead, through shared.

Yes, we haven't thought to much on how to make it optimal for standalone use. In the ASP.NET scenario it is recommended to leave the lifetime handling of HttpClient to the ASP.NET DI-system, and therefore that is what is needed in the constructor. For standalone usage, there might be overloads that could make things easier. Please have input in an Issue if you have any ideas :)

> We store our certificates in the windows certificate store so that's a nice alternative to the disk based, if you don't use azure. But that's a separate PR/issue if you are interested.

Yeah, a PR on that would be great!

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
ellinge: Modified the extensionmethods as well

ellinge:
> > First off. Thanks for the repo to start off with.
> 
> Thank you, I like how we can collaborate on things like these :) Out of interest, can you disclose what project you are using it for?
>
Yeah that's great! Was suprised the v4 endpoint was "already" stamped with end of life, but it's a rest world...
Not sure what the client thinks so I'll leave it to them. But basically, we've lifted the now soon deprecated v4-endpoint to v5 for their sign in (which we added on top of their password based "my pages"). Now we’re prototyping generation of dynamic agreements (pdfs) which is then signed by all parties involved. We investigated in existing services for this but decided a self-hosted solution would be best for the prototype, based on the requirements.
> > I only use the client-part. Something which would be nice is if some of the HttpClient setup parts where moved to the client-libraries instead, through shared.
> 
> Yes, we haven't thought to much on how to make it optimal for standalone use. In the ASP.NET scenario it is recommended to leave the lifetime handling of HttpClient to the ASP.NET DI-system, and therefore that is what is needed in the constructor. For standalone usage, there might be overloads that could make things easier. Please have input in an Issue if you have any ideas :)
>
Mostly thought that some of the BankIdAuthenticationBuilder configuration could be decoupled from AuthenticationBuilder/AspNetCore, and moved to the API-library, for standalone use. But not an issue since we need to configure it manually either way (window store) with our container.
> > We store our certificates in the windows certificate store so that's a nice alternative to the disk based, if you don't use azure. But that's a separate PR/issue if you are interested.
> 
> Yeah, a PR on that would be great!
>
I'll have that in mind then! Not sure yet if we can use the rest of the libraries or not based on other factors.

PeterOrneholm:
Fantastic @ellinge! I'm happy to merge this now :) Thank you so much for your contribution. I'll see if I can get a 1.1.0-alpha1 out on nuget any day now so you can start using that instead of your own.

fredrik-lundin:
This is awesome! Great job!

PeterOrneholm:
Short status update. We are finishing up a few other things and are planning on releasing this feature that you implemented in an upcoming 2.0.0 release. It will first be released with a `beta1` suffix so you can try it out before going live.

@ellinge: We would like to send you a small gift to say thank you for this valuable contribution (the first major external PR merged into Active Login!). Feel free to drop me an email (found [in profile](https://github.com/PeterOrneholm/)) with your details and I'll arrange it :)

PR #90, opened 2019-02-21
[WIP] Towards Helm chart

haf:
To consider: https://github.com/dotnet/dotnet-docker/blob/master/samples/dotnetapp/Dockerfile.alpine-x64-selfcontained

DoD:

 - All apps needed to run a full OAuth2/OpenId Connect flow have been dockerised
 - We have a [demo of the setup in skaffold](https://github.com/GoogleCloudPlatform/microservices-demo/tree/master/src)
 - We have a helm chart for using ActiveLogin as an IdP together with e.g. Ory Hydra or IdentityServer.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+*.Dockerfile
+Dockerfile
--- a/IdentityServer.ServerSample.Dockerfile
+++ b/IdentityServer.ServerSample.Dockerfile
@@ -8,10 +8,10 @@ RUN dotnet restore samples/IdentityServer.ServerSample/IdentityServer.ServerSamp
 RUN dotnet publish -c Release samples/IdentityServer.ServerSample/IdentityServer.ServerSample.csproj
 
 # ------------ RUN -----------
-FROM microsoft/dotnet:2.2-runtime
+FROM microsoft/dotnet:2.2-aspnetcore-runtime
 
 WORKDIR /app/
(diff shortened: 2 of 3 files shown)


haf:
img_6112

haf:
img_6113

PeterOrneholm:
Thanks for summarizing and uploading the docs from our fika, looking forward to make this happen! Can't promise when I/we'll find time, but I'll keep you posted. Will have a chat with Jakob.

haf:
If you can "rename" all samples/projects and make the top-level in this repo, I could take care of dockerising it all and creating the charts. Perhaps also add into this thread, together with the org of stuff you find good, a small paragraph on the actual order, so I know what projects to dockerise.

PeterOrneholm:
> If you can "rename" all samples/projects and make the top-level in this repo

Can you give me an example of this? I'd like to keep the current folder structure if possible, and would like to suffix the current samples as samples. Once they are dockerized we'll look at creating a slimmed down version that can be deployed out-of-the box. Preferably called something with template or similar.

haf:
Like `samples/IdentityServer` is not a good name for the ActiveLogin portal of the "default happy path"; there's nothing particular to identityserver in it and it's not a sample in this case.

PeterOrneholm:
Will close this for now. We will try to start of with a smaller sample, and Helm chart might come later.

PR #92, opened 2019-02-26
Update sample and documentation on partial view overriding

### LINKED ISSUE #79: Override for Razor Class Library does not work with current partial tag helpers
We use the latest `<partial name="" />` tag helper, but somehow it does not work when you override it in the actual MVC project, so the browser renders exactly the tag which means Razor does not recognise it as a valid partial wrapper.

One of the solutions could be to rollback to the old implementation of rendering partials, like `@await Html.PartialAsync("_partial.cshtml")`

Or we can completely get rid of the partials and start using components instead (https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-2.2). I like this option more, but needs to be tested if Razor engine on the top project can recognise it.

Zonnex:
Fixes #79.

Sample updated to show how to override partials
Documentation updated with text on how to override partials


samples/Standalone.MvcSample/Areas/BankIdAuthentication/Views/BankId/_Login.cshtml
```diff
@@ -0,0 +1,12 @@
+﻿@using Microsoft.Extensions.Localization
+@model ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthentication.Models.BankIdLoginViewModel
+@inject IStringLocalizer<ActiveLogin.Authentication.BankId.AspNetCore.BankIdAuthenticationHandler> Localizer
+@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
+
+<div class="bankid-login" style="width: 100%; max-width: 320px; margin: 0 auto; text-align: center; position: absolute; top: 50%; left: 50%; transform: translate(-50%, -50%);">
+    <div class="h3 mb-3 font-weight-normal">_Login partial from MvcSample</div>
```
PeterOrneholm: I'd like the sample still to use `Localizer` to show this usecase, maybe we can instead add a subheading or something similar?

README.md
```diff
@@ -141,6 +141,21 @@ For more use cases, samples and inspiration; feel free to browse our unit tests
 * [AzureProvisioningSample](https://github.com/ActiveLogin/ActiveLogin.Authentication/tree/master/samples/AzureProvisioningSample)
 * [ActiveLogin.Authentication.BankId.Api.Test](https://github.com/ActiveLogin/ActiveLogin.Authentication/tree/master/test/ActiveLogin.Authentication.BankId.Api.Test)
 
+## Overriding partial views
+
+ActiveLogin comes with predefined views that you can use, but maybe you'd rather use your own views.
```
PeterOrneholm: Great docs. Sugegsting adding this to the end:

```
.... but maybe you'd rather use your own views to customize layout or behavior.
```

README.md
```diff
@@ -141,6 +141,21 @@ For more use cases, samples and inspiration; feel free to browse our unit tests
...
+In your web project, create the following folder:
+```Areas/BankIdAuthentication/Views/BankId```
+In this folder, you can then create any of the partials
```
PeterOrneholm: I'd change this into:

In this folder, you can then create any of the partials and MVC will then discover your partials and use any of them before ours. It's still possible to call our partials if you still want to use them.

* `_Login.cshtml`
* `_LoginForm.cshtml`
* `_LoginScript.cshtml`
* `_LoginStatus.cshtml`

Zonnex:
I've updated the branch based on feedback. Ready to merge unless there's anything else.

Issue #99, opened 2019-03-04
Create an editorconfig file


Zonnex:
**Is your feature request related to a problem? Please describe.**
Considering we are open source, it seems to me like a good idea to include an .editorconfig file so it's easier for us and other contributors to adhere to coding style. I'm opening this issue to start the discussion regarding what conventions we want in our project. [Here](https://github.com/RehanSaeed/EditorConfig/blob/master/.editorconfig) is one we could start the discussions around. What do we like, what do we want to change?

**What area is it related to**
Source

**Describe the solution you'd like**
Include an EditorConfig in the solution

**Describe alternatives you've considered**
Resharper config. EditorConfig doesn't depend on Resharper, and there is a [VSCode extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig&WT.mc_id=-blog-scottha) available

**Additional context**

Zonnex:
Few things to note: [Existing files are not reformatted](https://github.com/editorconfig/editorconfig/wiki/FAQ#my-files-are-not-automatically-reformatted-the-editorconfig-plugin-is-not-working), so all changes the editorconfig would apply in new files are not shown in the draft PR. I think I'll run one of the tools mentioned [here](https://github.com/editorconfig/editorconfig/wiki/FAQ#are-there-any-tools-to-check-orand-reformat-existing-files-against-editorconfig-rules) to really show the changes.

Zonnex:
https://github.com/dotnet/roslyn/blob/master/.editorconfig

Here is the editorconfig used by roslyn.

Zonnex:
Sadly the new cool tool [dotnet format](https://www.hanselman.com/blog/CommentView.aspx?guid=29745397-2b1c-42c6-b1d6-f5002475c68b) won't apply all the rules, only [whitespace formatting](https://github.com/dotnet/format/issues/40#issuecomment-469304238) for the time being. I think it's a good idea to use ```suggestion``` rather than warning/error in the beginning unless someone wants to apply the rules manually across all our files. So for instance, we might decide to use coalesce expression ```var v = x ?? y``` everywhere, then we should initially use ```dotnet_style_coalesce_expression = true:suggestion``` so that we can gradually apply the rules as we come across discrepancies.

Zonnex:
https://github.com/Zonnex/EditorConfig

I uploaded my preferences based on my VS2019 settings.

PR #100, opened 2019-03-04
Use .editorconfig

Zonnex:
Formatted code with initial config mentioned in #99


.editorconfig
```diff
@@ -0,0 +1,283 @@
...
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset = utf-8
+end_of_line = crlf
+indent_style = space
+indent_size = 4
+insert_final_newline = false
```
fredrik-lundin: `false`? really..? really..!?

viktorvan: Agree, I would prefer **insert_final_newline = true** and **end_of_line = lf**

Zonnex: Why use one type of setting over another in this case?

Zonnex: https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file

These are the reasons?

Zonnex:
I've used [Resharper CodeCleanup CLI](https://blog.jetbrains.com/dotnet/2018/03/01/code-cleanup-resharper-command-line-tools/) to try apply rules to all files (since dotnet format only applies whitespace formatting).

One thing it does however is applying some spacing rules that that turns this. 
```new AuthenticationProperties { RedirectUri = "/" }```
into this
```new AuthenticationProperties {RedirectUri = "/"}```

I haven't found where the setting is yet, but I'll find it soon.
One more thing it does is reorder private fields, I'll see if I can disable that. I'd like to have it just run all rules defined in editorconfig but no success yet.

File: samples/IdentityServer.ClientSample/Startup.cs [hunk 2]
```diff
@@ -66,11 +67,8 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
             app.UseHttpsRedirection();
 
-            if (env.IsDevelopment())
-            {
-                app.UseDeveloperExceptionPage();
-            }
-            
+            if (env.IsDevelopment()) app.UseDeveloperExceptionPage();
```
PeterOrneholm: I usually don't like these on one line, what do you think?

samples/IdentityServer.ServerSample/Controllers/AccountController.cs
```diff
@@ -23,8 +24,8 @@ public class AccountController : Controller
 
         public IActionResult Login(string returnUrl)
         {
-            var schemes = _authenticationSchemeProvider.GetAllSchemesAsync();
-            var providers = schemes.Result
+            Task<IEnumerable<AuthenticationScheme>> schemes = _authenticationSchemeProvider.GetAllSchemesAsync();
```
PeterOrneholm: Hmm, i had missed this, but why are we not doing async/await?

samples/IdentityServer.ClientSample/Startup.cs
viktorvan: If it’s less than 120 chars wide I am fine with a one liner, but I still like to use braces.

Zonnex: Hi, thanks this one is fixed now.

samples/IdentityServer.ServerSample/Controllers/AccountController.cs
Zonnex: No idea actually! It should be

Zonnex:
Regarding var vs types. I currently try to follow the [corefx guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md). Thoughts?

.editorconfig
fredrik-lundin: Basically yes. But mostly it's just a cosmetic thing for me. I'm so used to seeing new lines at the end of a file, so when it's not there, it looks messy and compact.

Note that the editor config for roslyn (https://github.com/dotnet/roslyn/blob/master/.editorconfig) that you linked to in the related issue sets `insert_final_newline = true`. It's a very common thing to do.

samples/IdentityServer.ClientSample/Startup.cs
Zonnex: My bad, I missed the brace preference. I usually prefer single line statements/expressions to be without braces but we can put in preference for.

fredrik-lundin:
> I've used [Resharper CodeCleanup CLI](https://blog.jetbrains.com/dotnet/2018/03/01/code-cleanup-resharper-command-line-tools/) to try apply rules to all files (since dotnet format only applies whitespace formatting).
> 
> One thing it does however is applying some spacing rules that that turns this.
> `new AuthenticationProperties { RedirectUri = "/" }`
> into this
> `new AuthenticationProperties {RedirectUri = "/"}`
> 
> I haven't found where the setting is yet, but I'll find it soon.
> One more thing it does is reorder private fields, I'll see if I can disable that. I'd like to have it just run all rules defined in editorconfig but no success yet.

Have you tried playing around with `spaces_around_brackets`

Looks like it could be what you are looking for (https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties):
image

samples/IdentityServer.ServerSample/Startup.cs
```diff
@@ -108,47 +102,47 @@ public void ConfigureServices(IServiceCollection services)
...
+                        {
+                            options.IssueBirthdateClaim = true;
+                            options.IssueGenderClaim = true;
+                        })
+                        .AddBankIdSameDevice(GrandIdAuthenticationDefaults.BankIdSameDeviceAuthenticationScheme,
+                            "GrandID (SameDevice)", options => { })
+                        .AddBankIdOtherDevice(GrandIdAuthenticationDefaults.BankIdOtherDeviceAuthenticationScheme,
+                            "GrandID (OtherDevice)", options => { })
+                        .AddBankIdChooseDevice(GrandIdAuthenticationDefaults.BankIdChooseDeviceAuthenticationScheme,
+                            "GrandID (ChooseDevice)", options => { });
```
Zonnex: Bad wrapping here

samples/IdentityServer.ServerSample/Startup.cs
```diff
@@ -108,47 +102,47 @@ public void ConfigureServices(IServiceCollection services)
...
 
                     if (Configuration.GetValue("ActiveLogin:GrandId:UseSimulatedEnvironment", false))
-                    {
                         builder.UseSimulatedEnvironment();
-                    }
                     else
-                    {
                         builder.UseProductionEnvironment(config =>
```
Zonnex: Lack of braces here looks bad I think

samples/IdentityServer.ClientSample/Startup.cs
Zonnex: I have changed the editorconfig to have ```csharp_prefer_braces = true:suggestion```

fredrik-lundin:
Any reasons this PR never got completed? @PeterOrneholm @Zonnex

Zonnex:
@nikolaykrondev 
`We only use var when it's obvious what the variable type is (e.g. var stream = new FileStream(...) not var stream = OpenStandardInput()).`
The idea is that code reviewing outside an editor is slightly annoying if you can't easily see what the types are. Types makes the code easier to reason about.

Zonnex:
We recently did this in a few projects, and my suggestion is to generate a new editorconfig, recommit the changes and we'll review this again after

viktorvan:
Well nowadays you can just run gitpod.io on your repo, and get an IDE in the browser, so that is almost getting to be a moot point. 

https://gitpod.io/#https://github.com/ActiveLogin/ActiveLogin.Authentication

`var` isn’t just a shorter syntax, it does come with some other benefits.

PeterOrneholm:
This is so out of date with master, so I'll close it, and we can make a new one when implementing it.

PR #105, opened 2019-03-21
Cleanup code smells

### LINKED ISSUE #104: Fix codesmells found by SonarCloud
SonarCloud has found a few code smells (https://sonarcloud.io/dashboard?id=ActiveLogin_ActiveLogin.Authentication), I don't think all should be fixed as they might be opiniated, but some are low hanging fruits.

Fix these:
- `Add a 'protected' constructor or the 'static' keyword to the class declaration. `
- `'System.Exception should not be thrown by user code.`
- `Update this implementation of 'ISerializable' to conform to the recommended serialization pattern.`
- `Fix this implementation of 'IDisposable' to conform to the dispose pattern.`
- `Merge this if statement with the enclosing one.`
- `'TResult' is not used in the method.`
- `Use a constructor overloads that allows a more meaningful exception message to be provided.`

Maybe:
- `Remove this commented out code.` Can we handle the different scenarios in other ways?

Not:
- `Refactor your code not to use hardcoded absolute paths or URIs.` I'd say we should have these hard coded in this case, they  are constants. Maybe we should split the constant of URL and the URI object though?
- `Add a <legend> tag to this fieldset.` Legend is optional and does not bring value here.
- `Constructor has X parameters, which is greater than the 7 authorized.` Don't fix now, is breaking change.

PeterOrneholm:
Fixes #104.

### DIFF
--- a/samples/IdentityServer.ClientSample/Program.cs
+++ b/samples/IdentityServer.ClientSample/Program.cs
@@ -3,7 +3,7 @@ using Microsoft.AspNetCore.Hosting;
 
 namespace IdentityServer.ClientSample
 {
-    public class Program
+    public static class Program
     {
         public static void Main(string[] args)
         {
--- a/samples/IdentityServer.ServerSample/Config.cs
+++ b/samples/IdentityServer.ServerSample/Config.cs
@@ -6,7 +6,7 @@ using Microsoft.Extensions.Configuration;
 
 namespace IdentityServer.ServerSample
 {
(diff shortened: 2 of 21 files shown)


PeterOrneholm:
Fixes everything except:

- `'System.Exception should not be thrown by user code.`
- `Update this implementation of 'ISerializable' to conform to the recommended serialization pattern.`

Should we fix them? I'm not sure at this stage,

PR #107, opened 2019-05-20
 SPAR data in UserAttributes

krompaco:
New model and property for when UserAttributes in GrandId Api response contains SPAR data.

### DIFF
--- a/src/ActiveLogin.Authentication.GrandId.Api/Models/BankIdGetSessionUserAttributes.cs
+++ b/src/ActiveLogin.Authentication.GrandId.Api/Models/BankIdGetSessionUserAttributes.cs
@@ -17,6 +17,9 @@ namespace ActiveLogin.Authentication.GrandId.Api.Models
             IpAddress = ipAddress;
         }
 
+        [DataMember(Name = "SPAR")]
+        public BankIdGetSessionUserAddressAttributes AddressAttributes { get; set; }
+
         [DataMember(Name = "signature")]
         public string Signature { get; private set; }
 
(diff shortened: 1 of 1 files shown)


PeterOrneholm:
Great work @krompaco! Thanks for your contibution!

Before I start reviewing the code in detail, do you have some documentation available for this?
I haven't used SPAR myself, so would be good to have something to check against.

For example, when are these returned? is it enabled on your "license"? Do you need to request it with some param? Etc.

Also, I'd like to see a few tests (see our existing tests as samples) and also implement this in the simulated environment client.

Also, I might have a few comments on the naming. I think I'd like the properties to contain the word SPAR to be clear where the data comes from.

And also, i'd like to have extension methods for the data that can be parsed out of those strings. For example, the dates now represented as strings, can be parsed as `DateTime` See an [example here](https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/master/src/ActiveLogin.Authentication.GrandId.Api/Models/BankIdGetSessionUserAttributesExtensions.cs).

krompaco:
Thanks!

I was thinking this was more of proof-of-concept commit and was fairly certain you would have input :)

It's automatic and activated by the supplier. No parameter is needed from the client side.

I just found the property spakoUtdelningsadress1 can show up as well if it's a really long address.

They also sent me this documentation link but there seems to be no docs on the actual GrandId setup.

https://www.statenspersonadressregister.se/download/18.36e6b65d16141e753a6e12/1527515059771/SPAR_Systemgr%C3%A4nssnitt_ver_2-06.pdf

PeterOrneholm:
Thanks for the update :) Glad you are contributing to this project!

> It's automatic and activated by the supplier. No parameter is needed from the client side.

I see, convenient in one way, but I guess it might add unintended cost? (I haven't used it, so don't know the pricing model.) For the simulated environment, I'd say we should have a parameter specifying if the SPAR-data should be provided or not, and it should be false by default.

> They also sent me this documentation link but there seems to be no docs on the actual GrandId setup.

Thanks! Do they use the exact same property names etc?

I'd be happy to add the SPAR-support if you add the tests, simulated environment support, extensions, documentation etc mentioned above :)

PeterOrneholm:
Hi again @krompaco!
Doing some cleaning up in this repo, do you think you will continue working on this PR or should I close it for now? Let me know if you want to discuss the implementation.
/Peter

PeterOrneholm:
Closing this for now, with reason in comments above.

Issue #123, opened 2019-10-09
Support callback path based on IApplicationBuilder.Map paths


span:
**Is your feature request related to a problem? Please describe.**
The current callback path does not work together with path matches defined in `IApplicationBuilder.Map`. If adding the path match prefix from `IApplicationBuilder.Map` to the callback path, the callback endpoint becomes nested.

**What area is it related to**
BankIdAuthenticationHandler

**Describe the solution you'd like**
An option to add the path that I use in `IApplicationBuilder.Map` to the callback path. This can then be used in an override of `ShouldHandleRequestAsync` to handle the nested path.

**Describe alternatives you've considered**
URL rewriting etc.

**Additional context**
I have an open stack overflow question on how to work around the issue but it does not seem possible without changing the `BankIdAuthenticationHandler`.

https://stackoverflow.com/questions/58249177/using-remoteauthenticationhandler-callbackpath-with-iapplicationbuilder-path-mat?noredirect=1#comment102983312_58249177

I could spend the time implementing this if you think it it necessary and will make for a good addition.

fredrik-lundin:
Hi,

Thanks for posting such a detailed and well explained issue. I definitely think your scenario is something that we should support. Even though it might not be the most common use-case to wrap whole routing inside a `map`, it's certainly not an extreme edge case in my eyes.

The answer you've got in your stack overflow thread looks promising to me (https://stackoverflow.com/a/58317651/2521893). It points out that the _matched path segments_ is added to `HttpRequest.PathBase` when running inside the context of a `map` . 

From the asp.net core docs (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-3.0#use-run-and-map):

> When Map is used, the matched path segments are removed from HttpRequest.Path and appended to HttpRequest.PathBase for each request.

So a possible solution could probably be to register a `IHttpContextAccessor`, then use that inside to `RemoteAuthenticationHandler` and then do something like what is suggested in the Stack overflow answer:

```
HttpContext context = _httpContextAccessor.HttpContext;
string returnUrl = UrlEncoder.Encode(context.Request.PathBase + Options.CallbackPath);
```

Nothing would make us happier than if you'd like to contribute with a pull request yourself! Either go with something like suggested above or if you have a better solution in mind feel free to try it out. I will gladly help out in discussion around this if needed!

span:
Very nice, I will dig right into it.

[...]

PR #128, opened 2019-10-19
Do not add path base in BankID api controller

span:
In commit 0fa6d0a2cb7b9552ebe58dbf57f8f55dc7f0dd10 path base handling
was added to the BankID authentication handler. It created a bug
where the redirect uri for auto launching the BankID app would
contain duplicate/nested path bases.

This commit removes the pathbase from the api controller.

Ref ActiveLogin/ActiveLogin.Authentication#123

### DIFF
--- a/test/ActiveLogin.Authentication.BankId.AspNetCore.Test/BankId_Authentication_Tests.cs
+++ b/test/ActiveLogin.Authentication.BankId.AspNetCore.Test/BankId_Authentication_Tests.cs
@@ -197,7 +197,7 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Test
 	        var loginResponse = await client.GetAsync("/BankIdAuthentication/Login?returnUrl=%2F&loginOptions=X&orderRef=Y");
 	        var loginCookies = loginResponse.Headers.GetValues("set-cookie");
 	        var loginContent = await loginResponse.Content.ReadAsStringAsync();
-	        var csrfToken = TokenHelper.ExtractToken(loginContent);
+	        var csrfToken = TokenExtractor.ExtractRequestVerificationTokenFromForm(loginContent);
 
 	        // Arrange acting request
 	        var testReturnUrl = "/TestReturnUrl";
@@ -246,7 +246,7 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Test
         {
             return app =>
             {
-                app.UseMiddleware<FakeRemoteIpAddressMiddleware>();
+                app.UseMiddleware<FakeRemoteIpAddressMiddleware>(IPAddress.Parse("192.0.2.1"));
                 app.UseAuthentication();
                 app.UseRouting();
                 app.UseEndpoints(endpoints =>
--- a/test/ActiveLogin.Authentication.BankId.AspNetCore.Test/BankId_Authentication_Tests.cs
+++ b/test/ActiveLogin.Authentication.BankId.AspNetCore.Test/BankId_Authentication_Tests.cs
@@ -15,7 +15,6 @@ using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Authorization;
-using Microsoft.Extensions.WebEncoders.Testing;
 using Moq;
 using Xunit;
 
@@ -42,9 +41,9 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Test
                     o.UseSimulatedEnvironment()
                      .AddSameDevice();
                 },
-	            DefaultAppConfiguration(async context =>
+                DefaultAppConfiguration(async context =>
                 {
-	                await context.ChallengeAsync(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme);
+                    await context.ChallengeAsync(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme);
                 })).CreateClient();
 
             // Act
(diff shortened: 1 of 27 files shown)


PeterOrneholm:
Thank you for this! Could you write a simple test for it? That would ensure we don't hit it again.

span:
I am working on a test at the moment but I find it pretty much impossible to isolate the code in question without refactoring. Instead I am hacking around sending csrf tokens etc. back and forth to be able to make sure the correct url is resolved.

I will push the test once it is complete for you to have a look at but currently I am not sure on what it will test  in the end :P

span:
So, I pushed a commit with a test for the resolving of the URI. It has a few helper classes that I needed to be able to step through the controller logic. Hopefully you do not find it too messy.

Testing was made difficult since most classes are `internal`, of course. Perhaps it would be useful to add an assembly configuration so the test project may access the internal classes under test.

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/BankId_Authentication_Tests.cs
```diff
@@ -161,7 +166,60 @@ public async Task BankIdAuthentication_Login_Returns_Form()
             Assert.Contains("<form id=\"bankIdLoginForm\">", content);
         }
 
+        [Fact]
+        public async Task AutoLaunch_Sets_Correct_RedirectUri()
+        {
+	        // Arrange mocks
+	        var identityNumber = "195008212226";
```
PeterOrneholm: Nice that you use a test PIN :)
My site says it's valid to use in code ;)

https://swedish.identityinfo.net/personalidentitynumber/validate?number=195008212226

PeterOrneholm:
I will continue to review this later during the day :) Looks good so far!

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/BankId_Authentication_Tests.cs
span: Yes, this is from skatteverket open data since using "real" numbers would cause all kinds of issues with privacy laws.

PeterOrneholm: Exactly :) The site I linked to is absed on that dataset. I created it bacause so many people are using real PINs in their code...

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs
```diff
@@ -0,0 +1,23 @@
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace ActiveLogin.Authentication.BankId.AspNetCore.Test.Helpers
+{
+    internal class FakeRemoteIpAddressMiddleware
+    {
+        private readonly RequestDelegate next;
```
PeterOrneholm: Resharper tells me that suggested name is `_name` (underscore prefix).

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs
```diff
@@ -0,0 +1,23 @@
...
+        private readonly IPAddress fakeAddress = IPAddress.Parse("10.0.0.1");
```
PeterOrneholm: Same here, underscore prefix: `_fakeAddress`.

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public FakeRemoteIpAddressMiddleware(RequestDelegate next)
```
PeterOrneholm: I would propose to take the IPAddress as a parameter, then it will be very clear form the what what IP is used.
I had to read up a bit, but here is how it can be used.

Final code would be:
```csharp
    internal class FakeRemoteIpAddressMiddleware
    {
        private readonly RequestDelegate _next;
        private readonly IPAddress _fakeAddress;

        public FakeRemoteIpAddressMiddleware(RequestDelegate next, IPAddress fakeAddress)
        {
            _next = next;
            _fakeAddress = fakeAddress;
        }
...
}
```

And usage is:
```csharp
app.UseMiddleware<FakeRemoteIpAddressMiddleware>(IPAddress.Parse("192.0.2.1"));
```

Also, tried reading up one what IP to use for test. ``192.0.2.X` seems to be a good range :)
https://stackoverflow.com/a/10456065

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/TokenHelper.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Text.RegularExpressions;
+
+namespace ActiveLogin.Authentication.BankId.AspNetCore.Test.Helpers
+{
+    public static class TokenHelper
```
PeterOrneholm: Small thing. Would propose naming it to `TokenExtractor` and make it internal.

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/TokenHelper.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public static string ExtractToken(string responseText)
```
PeterOrneholm: Suggestion to rename it to `ExtractRequestVerificationTokenFromForm`.

span: Absolutely!

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/TokenHelper.cs
span: I love small things :D

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs
span: Nice! I added a comment as well in the code to refer to the RFC.

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs
span: Nice catch.

test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs
span: yup!

span:
It seems my force push is not playing nice with the merges from master into the feature branch.

This is the latest commit: https://github.com/ActiveLogin/ActiveLogin.Authentication/commit/2f854b22787768760cd711d7c92b011f00f7b133

Issue #131, opened 2019-10-20
IdentityServer example not starting properly


span:
**Describe the bug**
After updating to .net core 3.0 and new IdentityServer packages the example shows an internal server error.

```
An unhandled exception occurred while processing the request.
InvalidOperationException: Endpoint IdentityServer.ServerSample.Controllers.HomeController.Index (IdentityServer.ServerSample) contains authorization metadata, but a middleware was not found that supports authorization.
Configure your application startup by adding app.UseAuthorization() inside the call to Configure(..) in the application startup code.
Microsoft.AspNetCore.Routing.EndpointMiddleware.ThrowMissingAuthMiddlewareException(Endpoint endpoint)

Stack Query Cookies Headers Routing
InvalidOperationException: Endpoint IdentityServer.ServerSample.Controllers.HomeController.Index (IdentityServer.ServerSample) contains authorization metadata, but a middleware was not found that supports authorization. Configure your application startup by adding app.UseAuthorization() inside the call to Configure(..) in the application startup code.
Microsoft.AspNetCore.Routing.EndpointMiddleware.ThrowMissingAuthMiddlewareException(Endpoint endpoint)
Microsoft.AspNetCore.Routing.EndpointMiddleware.Invoke(HttpContext httpContext)
Microsoft.AspNetCore.Localization.RequestLocalizationMiddleware.Invoke(HttpContext context)
IdentityServer4.Hosting.IdentityServerMiddleware.Invoke(HttpContext context, IEndpointRouter router, IUserSession session, IEventService events)
IdentityServer4.Hosting.MutualTlsTokenEndpointMiddleware.Invoke(HttpContext context, IAuthenticationSchemeProvider schemes)
Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
IdentityServer4.Hosting.BaseUrlMiddleware.Invoke(HttpContext context)
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

**What area is it related to**
Examples

**To Reproduce**
1. Launch IdentityServer server example
2. View exception in browser

**Expected behavior**
Authentication options displayed

**Screenshots**
![image](https://user-images.githubusercontent.com/502136/67166154-1b91f780-f38d-11e9-9d27-675a0365b22e.png)

**NuGet package version**
source from master branch

**Runtime version**
.net core 3.0

**Additional context**
[...]

PR #132, opened 2019-10-20
Add use authorization to support IdentityServer controller

span:
The identity server example was broken by 3.0 update. This fixes the issue by adding authorization to the pipeline since it is required by identityserver.

Ref #131


PeterOrneholm:
Shame on me for missing this. Thanks @span!

Issue #119, opened 2019-09-27
Add UI support for aborting a login


PeterOrneholm:
**Is your feature request related to a problem? Please describe.**
BankID does support a "Cancel"-feature, which we on an API-level, have support for. But at the moment, we don't expose any UI for that so a user can't abort an ongoing login.

**What area is it related to**
BankID

**Describe the solution you'd like**
Implement a cancel button during the login flow.

PeterOrneholm:
Closed by #134

PR #134, opened 2019-10-21
Add cancel button to abort BankID auth requests

span:
This commit adds a cancel button to the status view. When clicked the button triggers a post request to a new endpoint in the BankIdApiController. The new endpoint then asks the BankID client to cancel the outstanding authentication request.

If the cancellation is successful, control is returned to the GUI where the status information is hidden and the form field and buttons for initiating a BankID authn request are once again visible.

If an error occurs, the status of the error is displayed as in the check status call. Let's open a new feature request to add/redirect the user to something more meaningful: #135 .

Ref ActiveLogin/ActiveLogin.Authentication#119

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
@@ -262,5 +262,17 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthenticatio
             var delimiter = url.Contains("?") ? "&" : "?";
             return $"{url}{delimiter}{queryString}";
         }
+
+        [ValidateAntiForgeryToken]
+        [HttpPost("Cancel")]
+        public async Task<CancelResponse> Cancel(BankIdCancelApiStatusRequest request)
+        {
+            var orderRef = _orderRefProtector.Unprotect(request.OrderRef);
+            var result = await _bankIdApiClient.CancelAsync(orderRef.OrderRef);
+
+            _logger.BankIdAuthCancelled(orderRef.OrderRef);
+
+            return result;
+        }
     }
 }
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
@@ -265,14 +265,15 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthenticatio
 
         [ValidateAntiForgeryToken]
         [HttpPost("Cancel")]
-        public async Task<CancelResponse> Cancel(BankIdCancelApiStatusRequest request)
+        public async Task<ActionResult> Cancel(BankIdLoginApiCancelRequest request)
         {
             var orderRef = _orderRefProtector.Unprotect(request.OrderRef);
-            var result = await _bankIdApiClient.CancelAsync(orderRef.OrderRef);
+
+            await _bankIdApiClient.CancelAsync(orderRef.OrderRef);
 
             _logger.BankIdAuthCancelled(orderRef.OrderRef);
 
-            return result;
+            return Ok(BankIdLoginApiCancelResponse.Cancelled());
         }
     }
 }
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdCancelApiStatusRequest.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdCancelApiStatusRequest.cs
@@ -1,0 +1,16 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthentication.Models
+{
+    public class BankIdCancelApiStatusRequest
+    {
+        internal BankIdCancelApiStatusRequest()
+        {
+
+        }
+
+        [Required]
+        public string OrderRef { get; set; }
+
+    }
+}
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
@@ -121,11 +121,11 @@
                             checkStatus(requestVerificationToken, returnUrl, loginOptions, data.orderRef);
                         }
 
-                        cancelButtonElement.addEventListener('click', function (event) {
-                            event.preventDefault();
+                        var onCancelButtonClick = function(event) {
                             cancel(requestVerificationToken, data.orderRef);
-                            cancelButtonElement.onclick = undefined;
-                        });
+                            event.target.removeEventListener('click', onCancelButtonClick);
+                        };
+                        cancelButtonElement.addEventListener('click', onCancelButtonClick);
                     }
                     show(formSubmitButtonTextElement);
                     hide(formSubmitButtonSpinnerElement);
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
@@ -80,7 +80,7 @@
 
         // BankID
 
-        var loginIsCancelledByUser;
+        var loginIsCancelledByUser = false;
         function initialize(requestVerificationToken, returnUrl, loginOptions, personalIdentityNumber, personalIdentityNumberElement) {
             loginIsCancelledByUser = false;
             formFieldsetElement.disabled = true;
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
@@ -4,8 +4,10 @@
     (function (options) {
         // Pre check
         var statusElement = document.querySelector('#bankIdLoginStatus');
+        var statusCardElement = statusElement.querySelector('.card');
         var statusSpinnerElement = statusElement.querySelector('.status-spinner');
         var statusMessageElement = statusElement.querySelector('.status-message');
+        var cancelButtonElement = statusElement.querySelector('.btn-cancel');
 
         var requiredFeatures = [window.fetch, window.sessionStorage];
         var isMissingSomeFeature = requiredFeatures.some(function(x) { return !x; });
@@ -78,7 +80,9 @@
 
         // BankID
 
+        var cancelled;
         function initialize(requestVerificationToken, returnUrl, loginOptions, personalIdentityNumber, personalIdentityNumberElement) {
+            cancelled = false;
             formFieldsetElement.disabled = true;
             show(formSubmitButtonSpinnerElement);
             hide(formSubmitButtonTextElement);
@@ -116,6 +120,12 @@
                         if (data.checkStatus) {
                             checkStatus(requestVerificationToken, returnUrl, loginOptions, data.orderRef);
                         }
+
+                        cancelButtonElement.onclick = function(e) {
+                            e.preventDefault();
+                            cancel(requestVerificationToken, returnUrl, loginOptions, data.orderRef);
+                            cancelButtonElement.onclick = undefined;
+                        };
                     }
                     show(formSubmitButtonTextElement);
                     hide(formSubmitButtonSpinnerElement);
@@ -126,6 +136,21 @@
                     hide(formSubmitButtonSpinnerElement);
                     formFieldsetElement.disabled = false;
                 });
+        }
+
+        function cancel(requestVerificationToken, returnUrl, loginOptions, orderRef) {
+            cancelled = true;
+            postJson(options.bankIdCancelApiUrl, requestVerificationToken, {
+                'orderRef': orderRef
+            }).then(function() {
+                hide(statusElement);
+                show(formElement);
+                formFieldsetElement.disabled = false;
+            }).catch(function(error) {
+                showStatus(error.message, 'danger', false);
+                hide(startBankIdAppButtonElement);
+                hide(cancelButtonElement);
+            });
         }
 
         function checkStatus(requestVerificationToken, returnUrl, loginOptions, orderRef) {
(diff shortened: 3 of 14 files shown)


PeterOrneholm:
You are on fire 🔥 ! :D This is a fantastic addition, thanks for implementing it.
Will add it to the que of things for me to review ;) Will have more time later this week.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -4,8 +4,10 @@
     (function (options) {
         // Pre check
         var statusElement = document.querySelector('#bankIdLoginStatus');
+        var statusCardElement = statusElement.querySelector('.card');
```
PeterOrneholm: As the class `card` comes from Bootstrap, I would like to have something more specific for this.
Could you add `status-wrapper` as a class as well? That gives us something unique to look for.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -78,7 +80,9 @@
 
         // BankID
 
+        var cancelled;
```
PeterOrneholm: Propse chaning this to: `loginIsCancelledByUser` to make it more clear what kind of cancel it is :)

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -116,6 +120,12 @@
                         if (data.checkStatus) {
                             checkStatus(requestVerificationToken, returnUrl, loginOptions, data.orderRef);
                         }
+
+                        cancelButtonElement.onclick = function(e) {
```
PeterOrneholm: Use `addEventListener` for listetning to JS events, like this:

```
            cancelButtonElement.addEventListener('click', function (event) {
                ......
            });
```

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -128,6 +138,21 @@
                 });
         }
 
+        function cancel(requestVerificationToken, returnUrl, loginOptions, orderRef) {
```
PeterOrneholm: These parameters are not used and can be removed:
- `returnUrl`
- `loginOptions`

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -139,12 +164,14 @@
                     if (data.isFinished) {
                         window.location.href = data.redirectUri;
                     } else {
+                        if (cancelled) return;
```
PeterOrneholm: I think we have used multi line if on other plac es, so:

```js
if (cancelled) {
   return;
};
```

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -139,12 +164,14 @@
...
                         showStatus(data.statusMessage, 'light', true);
                         setTimeout(function () {
                             checkStatus(requestVerificationToken, returnUrl, loginOptions, orderRef);
                         }, options.refreshIntervalMs);
                     }
                 }).catch(function (error) {
+                    if (cancelled) return;
```
PeterOrneholm: Same here then:

```js
if (cancelled) {
   return;
};
```

src/ActiveLogin.Authentication.BankId.AspNetCore/LoggerExtensions.cs
```diff
@@ -31,6 +31,11 @@ public static void BankIdAuthSuccess(this ILogger logger, SwedishPersonalIdentit
             logger.LogTrace(BankIdLoggingEvents.BankIdAuthSuccess, "BankID auth succedded for PersonalIdentityNumber '{PersonalIdentityNumber}' with the OrderRef '{OrderRef}'", personalIdentityNumber?.To12DigitString() ?? MissingPersonalIdentityNumber, orderRef);
         }
 
+        public static void BankIdAuthCancelled(this ILogger logger, string orderRef)
+        {
+            logger.LogInformation(BankIdLoggingEvents.BankIdAuthSuccess, "BankID auth was cancelled with the OrderRef '{OrderRef}'", orderRef);
```
PeterOrneholm: This is probably a copy paste mistake.

Would like to have the event like this in `BankIdLoggingEvents.cs`:
`public static readonly EventId BankIdAuthHardCancel = new EventId(2_1_3, nameof(BankIdAuthHardCancel));`

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginStatus.cshtml
```diff
@@ -2,13 +2,20 @@
...
+        <div class="card-body">
+            <div class="status-spinner">
+                <partial name="_spinner" />
+            </div>
+            <div class="status-message">@Model.LoginScriptOptions.InitialStatusMessage</div>
         </div>
-        <div class="status-message">@Model.LoginScriptOptions.InitialStatusMessage</div>
     </div>
+
+    <button type="button" class="btn btn-danger btn-block btn-cancel mt-2">
```
PeterOrneholm: I'm not 100% sure, but I think we should use `btn-link` instead of `btn-danger`. Danger feels so... dangerous :)
I had a look at other sites, like Handelsbanken, and they do the "Link" style.

With `.btn-link`:
A

With `.btn-danger`:
B

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
```diff
@@ -262,5 +262,17 @@ private static string AppendQueryString(string url, string queryString)
             var delimiter = url.Contains("?") ? "&" : "?";
             return $"{url}{delimiter}{queryString}";
         }
+
+        [ValidateAntiForgeryToken]
+        [HttpPost("Cancel")]
+        public async Task<CancelResponse> Cancel(BankIdCancelApiStatusRequest request)
```
PeterOrneholm: There should be a `BankIdLoginApiCancelResponse` to follow current conventions, even if only empty. Don't want to expose the API model to the client.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdCancelApiStatusRequest.cs
```diff
@@ -0,0 +1,16 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthentication.Models
+{
+    public class BankIdCancelApiStatusRequest
```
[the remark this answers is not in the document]
span: oops!

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginStatus.cshtml
span: I completely agree.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
span: Nice catch!

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Ok!

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Ok!

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Absolutely.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Yup, i'll try `else if(!loginIsCancelledByUser)` even. Then we wont need the return.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Same as above

src/ActiveLogin.Authentication.BankId.AspNetCore/LoggerExtensions.cs
span: Very nice, didn't see those.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Absoulutely. Love the decoupling it gives.

span:
Pushed an update with fixes. Would love a new review :)

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -116,6 +120,12 @@
                         if (data.checkStatus) {
                             checkStatus(requestVerificationToken, returnUrl, loginOptions, data.orderRef);
                         }
+
+                        cancelButtonElement.addEventListener('click', function (event) {
+                            event.preventDefault();
+                            cancel(requestVerificationToken, data.orderRef);
+                            cancelButtonElement.onclick = undefined;
```
PeterOrneholm: Realised this won't work now.
Would suggest:
```js
if(!loginIsCancelledByUser) {
    cancel(requestVerificationToken, data.orderRef);
}
```

Should have the same result, right?

PeterOrneholm:
The failing build seems to be a temp issue with macOS trying to download Core 3.0. Will rebuild once you push a commit for the thing above.

PeterOrneholm:
> Did you have a chances to look at a test for it? Can give that a try myself otherwise.

Started looking at this. Yeez, harder then I thought. So much is internal :)

span:
Yes. Unit testing is pretty much impossible so all that is left is system tests. It is made very difficult when things are internal and also there is no real "service" layer since much of the logic is embedded inside the controllers.

I think adding the tests as friend assemblies will make it much easier to test.

https://docs.microsoft.com/en-us/dotnet/standard/assembly/friend

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Hmm, I do not think I understand which context you are referring to.

PeterOrneholm:
Yes, we should have another look on how to improve the testing. I want to lock down as much as possible since it is a public package, end everything made public people can use and changing those will cause breaking changes.


Any way, I found a way to test it. I pushed it to your branch. Please have a look :)
It just verifies that cancel was called at least once.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
PeterOrneholm: `cancelButtonElement.onclick = undefined;`

Since we are using `cancelButtonElement.addEventListener('click'` setting `onclick ` to undefined wont work. There is the `once` option, but that's not supported in IE (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) so would like to avoid that.

Checking the `loginIsCancelledByUser` is probably the easiest to not cancel multiple times.

span: I see! I thought I saw the `onclick = undefined` elsewhere in the code. Perhaps I was mistaken. It should work as you say.

PeterOrneholm: Ahhh, yes, you are right. I will fix that in another PR.

PeterOrneholm:
If you fix the thing mentioned above (removing `.onclick = undefined` and use the if instead), I'm ready to merge :)

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
span: Ok, so a last shot.

What do you think about something like this:

```
var onCancel = function(event) {
    cancel(requestVerificationToken, data.orderRef);
    event.target.removeEventListener('click', onCancel);
};
cancelButtonElement.addEventListener('click', onCancel);
```

It is more verbose but perhaps it is more "javascripty"?

PeterOrneholm: Sounds great! Would just rename it to `onCancelButtonClick`

PeterOrneholm:
Yaay! In a few minutes you can try it out in our demo environment :D

PeterOrneholm:
I missed to test this scenario XD
https://github.com/ActiveLogin/ActiveLogin.Authentication/issues/143

Can try to fix in tonight.

Issue #124, opened 2019-10-10
Hacktober - Make a PR and get a t-shirt!


PeterOrneholm:
Active Login loves contributions from the community! We have seen initiatives from other companies like DigitalOcean and Microsoft to hand out t-shirts for OSS contributions during October. This year we are doing the same!

The first 10 PR:s from the community to be accepted during October will get a Active Login t-shirts! Exact design will be presented later :) Please see details below:

- The contribution should follow our [Contribution guidelines](https://github.com/ActiveLogin/ActiveLogin.Authentication#contribute). First and issue, then PR, write tests etc.
- You should follow our [Code of conduct](https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/master/CODE_OF_CONDUCT.md). 
- The PR should be merged before the end of October. We might approve some days into November as well... :)
- The applicable repos are:
  - https://github.com/ActiveLogin/ActiveLogin.Authentication
  - https://github.com/ActiveLogin/ActiveLogin.Identity
- We do accept PR:s for documentation and samples as well, but don't forget to write an issue first so we agree on that it should be fixed/implemented!
- Once merged, drop us an email (found in [our profile](https://github.com/ActiveLogin)) with a link to your PR and we'll get back with details on shirt size etc.

PR #137, opened 2019-10-24
Text update

Palpie:
Updated texts to the latest recommendations in BankID Relying Party Guidelines 3.2.2 (https://www.bankid.com/assets/bankid/rp/bankid-relying-party-guidelines-v3.2.2.pdf).

RFA17 is now RFA17A and RFA17B. Where the old RFA17 was used I have changed to RFA17A. RFA17B is for when using QR codes.


PeterOrneholm:
Looks great, thanks @Palpie!

Please see #124 for your free t-shirt :)

PR #142, opened 2019-10-28
Add and apply editorconfig

### LINKED ISSUE #99: Create an editorconfig file

Zonnex:
New PR to resolve #99.

Thoughts are welcome. This editorconfig is based on how my visual studio is configured.
Cleanup is done in VS2019 with the "Code Cleanup" feature that is now available. 

How to: Right click solution and hover "Analyze and Code Cleanup", configure your profile and then in the same menu hit "Run Code Cleanup" with desired profile.

Known issue is that I can't get the cleanup to ignore cleanup for methods, this means that all methods will have to be expression bodies, or method bodies. There is also the option to convert only one-liners to expression bodies, but that means all one-liners will become expression bodies.
If anyone has a strong preference here, please raise opinions.

I'm also considering making a case regarding tabs over spaces: https://dev.to/alexandersandberg/why-we-should-default-to-tabs-instead-of-spaces-for-an-accessible-first-environment-101f


Zonnex:
Feedback so far is to use var everywhere, unfortunately my visual studio didn't want to do this though.
Would be good if anyone could clone and try apply the editorconfig, see if they have better luck

span:
I applied it in RIder and had a few issues. It seems Rider does not like the line breaks of the `.editorconfig`. I would suggest splitting the chained definitions in to separate lines.

Rider also did not like the empty properties. Commenting them out did the trick :+1:

Very nice initiative.

PeterOrneholm:
Due to merge issues, we choosed to fix this in a separate PR (#180). But credit goes to you @Zonnex :)

Issue #127, opened 2019-10-16
Azure KeyVault certificate path is misleading


span:
**Describe the bug**
Azure key vault certificate client requests a secret and not a certificate. This means the object url that can be copied from keyvault is incorrect and requires manual handling.

**What area is it related to**
Azure keyvault

**To Reproduce**
1. Go to Keyvault Certificates
2. Copy certificate identifier
3. Configure client certificate with the identifier
4. Initiate an authentication request
5. After challenge, view this error: 

```
[2019-10-16 10:34:39 Error] Microsoft.AspNetCore.Server.Kestrel
Connection id "0HLQI2EKHAAJJ", Request id "0HLQI2EKHAAJJ:00000004": An unhandled exception was thrown by the application.
System.ArgumentException: Invalid ObjectIdentifier: https://[KEYVAULT]/certificates/[CERTIFICATE NAME]/[IDENTIFIER]. segment [1] should be 'secrets/', found 'certificates/'
   at Microsoft.Azure.KeyVault.ObjectIdentifier..ctor(String collection, String identifier)
   at Microsoft.Azure.KeyVault.SecretIdentifier..ctor(String identifier)
   at Microsoft.Azure.KeyVault.KeyVaultClientExtensions.GetSecretAsync(IKeyVaultClient operations, String secretIdentifier, CancellationToken cancellationToken)
   at ActiveLogin.Authentication.BankId.AspNetCore.Azure.KeyVault.AzureKeyVaultCertificateClient.GetX509Certificate2Async(String keyVaultSecretIdentifier) in [PROJECT PATH]/ActiveLogin.Authentication/src/ActiveLogin.Authentication.BankId.AspNetCore.Azure/KeyVault/AzureKeyVaultCertificateClient.cs:line 69
```

> segment [1] should be 'secrets/', found 'certificates/'

**Expected behavior**
I expect the action to work and not trigger an error

**NuGet package version**
Latest master

**Runtime version**
.net core 2.2

**Causing code**
The use of `GetSecretAsync` in `AzureKeyVaultCertificateClient` on line 69:

```
var secret = await _keyVaultClient
    .GetSecretAsync(keyVaultSecretIdentifier).ConfigureAwait(false);
```

**Proposed fix**
Use `GetCertificateAsync` to retreive certificates. Perhaps getting the certificate can be made smart so it accepts both `secrets` and `certificates`.

**Regressions**
[...]

PR #144, opened 2019-10-29
Support certificate identifier when loading from azure

span:
This commit adds support for loading a certificate identifier. It does
so by checking if the identifier is a 'secret' or 'certificate' in
the same manner that the 'SecretIdentifier' and 'CertificateIdentifier'
APIs in core do. By checking for a substring.

This commit does not remove support for fetching the certificates from
the 'secrets' in keyvault to avoid a breaking change.

Ref ActiveLogin/ActiveLogin.Authentication#127

The url checking seems to work fine but I would like to spend a bit more time testing.


src/ActiveLogin.Authentication.BankId.AspNetCore.Azure/KeyVault/AzureKeyVaultCertificateClient.cs
```diff
@@ -65,6 +66,21 @@ private AzureKeyVaultCertificateClient(KeyVaultClient client)
         }
 
         public async Task<X509Certificate2> GetX509Certificate2Async(string keyVaultSecretIdentifier)
```
PeterOrneholm: Would be nice to rename this to: `keyVaultIdentifier`, cause we support both secret and cert.
Also, we should then rename the other properties, for example in: `ClientCertificateFromAzureKeyVaultOptions`

This will cause a breaking change, but we can make next release a new major version. 
For `ClientCertificateFromAzureKeyVaultOptions.AzureKeyVaultSecretIdentifier` we would rename that, but create a (deprecated) set only prop with the old name, bot to break to many applications...

src/ActiveLogin.Authentication.BankId.AspNetCore.Azure/KeyVault/AzureKeyVaultCertificateClient.cs
```diff
@@ -77,6 +93,24 @@ public async Task<X509Certificate2> GetX509Certificate2Async(string keyVaultSecr
             return GetX509Certificate2(certificateBytes);
         }
 
+        private async Task<X509Certificate2> CertificateFromKeyVaultCertificate(string keyVaultSecretIdentifier)
```
PeterOrneholm: Probably a copy paste thing. `keyVaultSecretIdentifier` should be `keyVaultCertificateIdentifier`

span: Sorry for all these sloppy PR's. It seems my confidence is much better than my proof reading skills.

PeterOrneholm:
I've been trying this out now, and now I do remember the issue I ran into. I'll explain it below, and would be glad if we find a solution. I have one myself.

When I run this, I do get an exception because the cert is not exported with a private key. It fails here (line 119): `exportedCertCollection.Cast<X509Certificate2>().First(x => x.HasPrivateKey)`

According to this [SO-answer](https://stackoverflow.com/a/51482569), you need to access it as a secret to get it with the private key. Tried to send in the cert without private key, and then the SSL connection to BankID can't be established.

Did you get it working?

The URL:s are convention bases, so if we get `/certificates/` we could replace that with `/secrets/`. But optimally we would find a way to export it with the private key, but seems to be designed in a way so we can't do that. Would love to discuss the best way :)

src/ActiveLogin.Authentication.BankId.AspNetCore.Azure/KeyVault/AzureKeyVaultCertificateClient.cs
PeterOrneholm: Haha, na, no problem. That's why we do Code Reviews :) I'm the same myself!

span:
Wow, I had no idea. I read up on [the link ](https://stackoverflow.com/a/43839241/1068167)from the SO-answer and then also the very extensive description on [keyvault secrets and certificates](https://docs.microsoft.com/en-gb/azure/key-vault/about-keys-secrets-and-certificates#BKMK_CompositionOfCertificate).

It seems to me we could do a few things.

1. Scrap this whole PR and clarify in documentation on _why_ we need to use `/secrets/`.
2. Be convenient and replace `/certificates/` with secrets.
3. Be friendly and add a descriptive exception if the user provides `/certificates/` to make them change the url

Number 2 does hidden magic so I do not think I like it. 1 and 3 would be most user friendly I suppose.

span:
Also see link https://azidentity.azurewebsites.net/post/2018/07/03/azure-key-vault-certificates-are-secretsDen 2 nov. 2019 21:33 skrev Peter Örneholm <notifications@github.com>:I've been trying this out now, and now I do remember the issue I ran into. I'll explain it below, and would be glad if we find a solution. I have one myself.
When I run this, I do get an exception because the cert is not exported with a private key. It fails here (line 119): exportedCertCollection.Cast<X509Certificate2>().First(x => x.HasPrivateKey)
According to this SO-answer, you need to access it as a secret to get it with the private key. Tried to send in the cert without private key, and then the SSL connection to BankID can't be established.
Did you get it working?
The URL:s are convention bases, so if we get /certificates/ we could replace that with /secrets/. But optimally we would find a way to export it with the private key, but seems to be designed in a way so we can't do that. Would love to discuss the best way :)

PeterOrneholm:
> Number 2 does hidden magic so I do not think I like it. 1 and 3 would be most user friendly I suppose.

I just relized checking for /certificate/ could go wrong as well. If you call your secret *certificate* part of the url will be */certificate/*, so if we throw an error, we would need to search for something more specific.

Maybe start with alt. 1 for now?

span:
Yes. Will abondon this now and create a new issue for the documentation.

Issue #143, opened 2019-10-29
Cancel login on "Same Device" gives blank screen


PeterOrneholm:
**Describe the bug**
I obviously forgot to test this scenario, but if you cancel the login when choosing "Same Device" the page becomes empty, only showing the title "BankID".

**What area is it related to**
BankID

**To Reproduce**
1. Go to the start page
2. Click on "BankID (Same Device)"
3. Click "Cancel"
4. The page is cleared

**Expected behavior**
I've checked how Handelsbanken does it, and if you cancel on "Other device", they do like we do. Show the PIN input again. But when you cancel on "Same device", they take you back to the selection of schemes (Same / Other etc). I'd expect that to happen.

**Screenshots**
![image](https://user-images.githubusercontent.com/1662918/67752978-4555b900-fa34-11e9-8312-b6e297e70ef1.png)

**NuGet package version**
Latest code in master.

**Additional context**
Suggestion would be that the server response for Cancel would include a redirectUrl, if set, the client would redirect there. That redirectUri would point to the "initial" page.

span:
Is this even a valid operation? Perhaps it should not be supported and an error be thrown if this configuration is attempted. Looking at the guidlines they  say:

> If the BankID app is installed on the same device, but the BankID app cannot be automatically started, the user must provide their personal number and manually start the BankID app on the same device. RP:s should consider this use case as a fallback in case the automatic start fails.

So on the same device it SHOULD auto launch. But if it does not work, the user MUST provide the PIN and manually launch the app.

PeterOrneholm:
I got an email with your comment @span, but seems to be gone here.

I think it is a valid scenario. We have an orderRef and I've tried it, and BankID does accept that we cancel it. To try it out, I used my PC but didn't allow the BankID app to open, then I cancelled it.

![image](https://user-images.githubusercontent.com/1662918/67757704-98336e80-fa3c-11e9-867c-71e2834c7b86.png)

A scenario could be you choose "this device" but realize you actually wanted "other device" and want to cancel the first operation.

While on it, I realize we do get an error almost every time we cancel. That's because it tries to get one last status for that cancel, and the orderRef does not longer exists. I think we should check for cancel earlier, like this:

```js
        function checkStatus(requestVerificationToken, returnUrl, loginOptions, orderRef) {
            if (loginIsCancelledByUser) {
                return;
            }

......

        }
```

span:
[...]

PR #138, opened 2019-10-27
Use QR codes for other device authentication

span:
This commit adds support for QR codes. By default it is only used
for 'other device' authentication in line with the BankID
relying party guidelines.

It uses the auto login functionality since using a QR code follows
the same initialization steps as the auto launch. No PIN is
required and no PIN change is allowed.

A new dependency is added to create the QR codes. It is
a nuget package named 'QRCoder'[1].

Ref ActiveLogin/ActiveLogin.Authentication#19

[1] https://github.com/codebude/QRCoder/

PR #145, opened 2019-10-29
Bug/cancel same device

span:
Use cancel return url for samedevice auth

This commit adds a new property 'cancelReturnUrl' that is
used by the samedevice scheme to redirect the user back
to the initial page if they cancel an ongoing auth
process.

Currently it limits the cancelUrl to the samedevice
only but it is possible to allow the cancel url to be
set for the otherdevice scheme as well without too
much hassle.

Ref ActiveLogin/ActiveLogin.Authentication#143

Note. This will most probably conflict with #138 . We will have to rebase once the first is completed.

### DIFF
--- a/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
+++ b/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
+using ActiveLogin.Authentication.BankId.AspNetCore;
 using IdentityServer.ServerSample.Models;
 using IdentityServer4.Models;
 using IdentityServer4.Services;
@@ -50,6 +51,11 @@ namespace IdentityServer.ServerSample.Controllers
                     {"scheme", provider}
                 }
             };
+
+            if (provider.Equals(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme))
+            {
+                props.Items["cancelReturnUrl"] = returnUrl;
+            }
 
             return Challenge(props, provider);
         }
--- a/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
+++ b/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
@@ -52,11 +52,6 @@ namespace IdentityServer.ServerSample.Controllers
                 }
             };
 
-            if (provider.Equals(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme))
-            {
-                props.Items["cancelReturnUrl"] = returnUrl;
-            }
-
             return Challenge(props, provider);
         }
 
--- a/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
+++ b/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
@@ -47,10 +47,15 @@ namespace IdentityServer.ServerSample.Controllers
                 RedirectUri = Url.Action(nameof(ExternalLoginCallback)),
                 Items =
                 {
-                    {"returnUrl", returnUrl},
-                    {"scheme", provider}
+                    { "returnUrl", returnUrl },
+                    { "scheme", provider },
                 }
             };
+
+            if (provider.Equals(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme))
+            {
+                props.Items.Add("cancelReturnUrl", Url.ActionLink("Login", "Account", new { returnUrl }));
+            }
 
             return Challenge(props, provider);
         }
--- a/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
+++ b/samples/IdentityServer.ServerSample/Controllers/AccountController.cs
@@ -49,13 +49,9 @@ namespace IdentityServer.ServerSample.Controllers
                 {
                     { "returnUrl", returnUrl },
                     { "scheme", provider },
+                    { "cancelReturnUrl", Url.ActionLink("Login", "Account", new { returnUrl }) }
                 }
             };
-
-            if (provider.Equals(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme))
-            {
-                props.Items.Add("cancelReturnUrl", Url.ActionLink("Login", "Account", new { returnUrl }));
-            }
 
             return Challenge(props, provider);
         }
--- a/samples/Standalone.MvcSample/Controllers/AccountController.cs
+++ b/samples/Standalone.MvcSample/Controllers/AccountController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
+using ActiveLogin.Authentication.BankId.AspNetCore;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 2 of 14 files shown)


src/ActiveLogin.Authentication.BankId.AspNetCore/Serialization/BankIdLoginOptionsSerializer.cs
```diff
@@ -10,7 +10,7 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Serialization
 {
     internal class BankIdLoginOptionsSerializer : IDataSerializer<BankIdLoginOptions>
     {
-        private const int FormatVersion = 1;
+        private const int FormatVersion = 3;
```
span: Set to 3 since the QR change is using 2 at it seemed closest to merge.

span:
You do know that every time one merges from master to a feature branch a kitten dies ;)

span:
Latest commit fixes 3.

PeterOrneholm:
> You do know that every time one merges from master to a feature branch a kitten dies ;)

Sorry... :)

> Is this correct?

Yes, I agree.

Will try to review tomorrow!

src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationHandler.cs
```diff
@@ -166,6 +167,21 @@ bool TryGetPinString(out string s)
             return null;
         }
 
+        private string GetCancelUrl(AuthenticationProperties properties)
```
PeterOrneholm: `GetCancelReturnUrl` is probably more clear.

src/ActiveLogin.Authentication.BankId.AspNetCore/Models/BankIdLoginOptions.cs
```diff
@@ -31,5 +33,6 @@ public class BankIdLoginOptions
         public bool AllowBiometric { get; }
 
         public bool UseQrCode { get; set; }
+        public string CancelReturnUrl { get; }
```
PeterOrneholm: Add one more linebreak to align with the rest of the props.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -146,14 +147,19 @@
...
-                'orderRef': orderRef
-            }).then(function() {
+                'orderRef': orderRef,
+                'cancelReturnUrl': cancelReturnUrl
+            }).then(function(data) {
                 hide(statusElement);
                 show(formElement);
                 formFieldsetElement.disabled = false;
+
+                if (data.cancelReturnUrl) {
```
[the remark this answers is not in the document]
span: Nice catch. Never got used to this double bang. It seems it is getting more and more traction.

samples/Standalone.MvcSample/Controllers/AccountController.cs
```diff
@@ -50,6 +51,11 @@ public IActionResult ExternalLogin(string provider)
                 }
             };
 
+            if (provider.Equals(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme))
+            {
+                props.Items["cancelReturnUrl"] = "~/";
```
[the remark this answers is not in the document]
span: Actually I don't think we need it in the IdentityServer sample either. This was a leftover from a previous implementation where both samples needed it. The only reason I can think of on why we would override is if we want cancel to go back "home" after using PIN instead of QR in the same device scheme. 

Perhaps it is enough to put in a sample in the documentation?

src/ActiveLogin.Authentication.BankId.AspNetCore/BankIdAuthenticationHandler.cs
```diff
@@ -166,6 +167,23 @@ bool TryGetPinString(out string s)
             return null;
         }
 
+        private string GetCancelReturnUrl(AuthenticationProperties properties)
+        {
+            // Default to root if no return url is set
+            var cancelReturnUrl = properties.Items["cancelReturnUrl"] ??
```
PeterOrneholm: The test is failing here now. I think you need to check if the properties contains the value.

span: Doh.

span:
> I still think we need to set an explicit cancelReturnUrl in IdentityServer,
> The reason is that in a client/server scenario the returnUrl is stored in the query string.
> 
> For example: If you navigate to:
> http://al-samples-mvcclient.azurewebsites.net/
> 
> You will be redirected to:
> https://al-samples-identityserver.azurewebsites.net/Account/Login?ReturnUrl=%2Fconnect%2Fauthorize%2Fcallback%3Fc ....
> 
> That return url will be lost if we redirect to the root of identity server, so we need to set it explicitly. RIght?

Currently the cancellation URL is set to the `ReturnUrl` parameter if not specific `cancelReturnUrl` is passed. So in your scenario both the cancellation url and the return url would be set to `%2Fconnect%2Fauthorize%2Fcallback%3Fc ...`.

At least that is the intent and in my testing it seems to work.

PeterOrneholm:
> Currently the cancellation URL is set to the `ReturnUrl` parameter if the `cancelReturnUrl` is not set. So in your scenario both the cancellation url and the return url would be set to `%2Fconnect%2Fauthorize%2Fcallback%3Fc ...`.

Nah, what I mean is, if you press cancel, you would expect to get to the landing page in Identity Server. And for that to work properly, that page needs to have a query string param with the original returnUrl. Otherwise the login flow will not work. So:
`returnUrl` = `/connect/authorize/callback?c...`.
`cancelReturnUrl` = `/?returnUrl=`%2Fconnect%2Fauthorize%2Fcallback%3Fc ...``

Right? If we were to redirect all the way back to the original returnUrl, the user can't choose another auth scheme.

samples/Standalone.MvcSample/Controllers/AccountController.cs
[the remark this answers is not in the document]
span: You are correct of course!

span:
Oh yes, I see it now. I was getting tricked by the root redirect in the sample!

samples/IdentityServer.ServerSample/Controllers/AccountController.cs
```diff
@@ -46,11 +47,16 @@ public IActionResult ExternalLogin(string provider, string returnUrl)
...
                 Items =
                 {
-                    {"returnUrl", returnUrl},
-                    {"scheme", provider}
+                    { "returnUrl", returnUrl },
+                    { "scheme", provider },
                 }
             };
 
+            if (provider.Equals(BankIdAuthenticationDefaults.SameDeviceAuthenticationScheme))
```
PeterOrneholm: I would always set this (`{ "cancelReturnUrl", .... },`), makes it cleaner and you do do the check later on wheter or not to use it.

PeterOrneholm:
Merged now, awesome work @span!

Two left :)

span:
w00t w00t! Awesome reviewing @PeterOrneholm !

PeterOrneholm:
> w00t w00t! Awesome reviewing @PeterOrneholm !

I like the conversations we have :)

PR #146, opened 2019-10-31
Use c# 8 nullable references

nikolaykrondev:


PeterOrneholm:
We focus on the one in #163

Issue #148, opened 2019-11-04
Clarify in documentation on why we need to use /secrets/ in when fetching certificates


span:
When we fetch client certificates from key vault we are using the `GetSecretAsync` method as described in #144 .

We should clarify why we need to do it this way in the documentation.

A few links: 

Official documentation (very extensive)
https://docs.microsoft.com/en-gb/azure/key-vault/about-keys-secrets-and-certificates#BKMK_CompositionOfCertificate

Problem solved in blog:
https://azidentity.azurewebsites.net/post/2018/07/03/azure-key-vault-certificates-are-secrets

PeterOrneholm:
Fixed in #149

PR #149, opened 2019-11-04
Clarify why we fetch certificates as secrets

span:
This was discussed extensively in ActiveLogin/ActiveLogin.Authentication#144. Basically
certificates with private keys are inly exposed as _secrets_ in keyvalt. The
_certificates_ endpoint will only return the crt.

Ref ActiveLogin/ActiveLogin.Authentication#148


PeterOrneholm:
Good, thanks!

Issue #152, opened 2019-11-05
Support forwarded for header


span:
**Is your feature request related to a problem? Please describe.**
When running behind reverse proxy or CDN the user ip adress is not resolved properly. It is only resolved through `HttpContext.Connection.RemoteIpAddress` without checking the forward headers.

**What area is it related to**
BankID

**Describe the solution you'd like**
Add support to check for forwarded headers that is on by default.

**Describe alternatives you've considered**
The support could be off by default to keep full backwards compatability but looking forward it seems turning it on by default is preferred.

span:
Closing since it is preferred the user configure their pipeline instead.

PR #153, opened 2019-11-05
Support forwarded and cdn headers

span:
This commit adds support for using the forwarded headers
as the users remote ip addess. This is useful when the
application is running behind a reverse proxy or in a cdn.

Ref ActiveLogin/ActiveLogin.Authentication#152

### DIFF
--- a/docs/getting-started-bankid.md
+++ b/docs/getting-started-bankid.md
@@ -226,6 +226,9 @@ BankId options allows you to set and override some options such as these.
     // Turn off 
     code and use personal identity number instead
     options.BankIdUseQrCode = false;
+
+    // Disable forwarded headers
+    options.AllowForwardedHeaders = false;
 });
 ```
 
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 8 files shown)


PeterOrneholm:
Thanks for bringing this up!

Just tried to google on the subject a bit. Found out about the `UseForwardedHeaders()` middleware:
https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.0

By using that, you should be able to accomplish the same way, but in a more generic way, and could configure alot of things, like what headers to look for.

Don't you think that would be a better solution? Maybe a comment in the readme about it?

span:
I am using that middleware running in a docker container, inside kubernetes (AKS) behind a nginx reverse proxy behind Cloudflare. The forwarded headers works by setting the headers I added to this commit.

The `HttpContext.Connection.RemoteIpAddress` shows an internal AKS IP for me.

PeterOrneholm:
> The middleware is clunky to configure and requires custom headers which _can_ be environment specific. To me it seems like a good addition to handle the standard "non-standard"-forwarded headers :P
> 
> That said. If you do not think it belongs here I'll go configure my middleware instead :)

I see your point, but would really like to keep such details out of the package.
I would say, give that middleware another round.

IF we were to do anything in regards to the IP, I would say we should make it an interface (like `IBankIdEndUserIpExtractor`), but maybe overkill?

I think you did a nice job with [FakeRemoteIpAddressMiddleware](https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/master/test/ActiveLogin.Authentication.BankId.AspNetCore.Test/Helpers/FakeRemoteIpAddressMiddleware.cs) for our tests. Couldn't a similar Middleware solve this issue?

span:
Yup. Closing this for now.

PR #154, opened 2019-11-06
Add retries for StartFailed hints

span:
This commit adds a retry mechanism for error results with
hint StartFailed. It will retry 5 times before giving up.

Lets see where we can take this naive approach.

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiErrorResponse.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiErrorResponse.cs
@@ -12,11 +12,11 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthenticatio
         public BankIdLoginApiErrorResponse(string errorMessage, CollectHintCode hintCode)
         {
             ErrorMessage = errorMessage;
-            HintCode = hintCode.ToString();
+            Retry = CollectHintCode.StartFailed.Equals(hintCode);
         }
 
         public string ErrorMessage { get; }
 
-        public string HintCode { get; }
+        public bool Retry { get; }
     }
 }
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiErrorResponse.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiErrorResponse.cs
@@ -9,14 +9,14 @@ namespace ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthenticatio
             ErrorMessage = errorMessage;
         }
 
-        public BankIdLoginApiErrorResponse(string errorMessage, bool retry)
+        public BankIdLoginApiErrorResponse(string errorMessage, bool retryLogin)
         {
             ErrorMessage = errorMessage;
-            Retry = retry;
+            RetryLogin = retryLogin;
         }
 
         public string ErrorMessage { get; }
 
-        public bool Retry { get; }
+        public bool RetryLogin { get; }
     }
 }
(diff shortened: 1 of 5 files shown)


src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Views/BankId/_LoginScript.cshtml
```diff
@@ -210,7 +211,14 @@
                     throw Error(options.unknownErrorMessage);
                 })
                 .then(function (data) {
-                    if (!!data.errorMessage) {
+                    if (data.hintCode === 'StartFailed') {
```
PeterOrneholm: I would move this logic to the server. Then it's easier to extend.
Return a bool, something like `retry`. And check the hint code in the backend instead.

span: Wouldn't that mean moving state to the server? So when there are multiple nodes behind a load balancer they will not know how many times they retried already?

span: Good idea!

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiErrorResponse.cs
```diff
@@ -7,6 +9,14 @@ internal BankIdLoginApiErrorResponse(string errorMessage)
             ErrorMessage = errorMessage;
         }
 
+        public BankIdLoginApiErrorResponse(string errorMessage, CollectHintCode hintCode)
+        {
+            ErrorMessage = errorMessage;
+            Retry = CollectHintCode.StartFailed.Equals(hintCode);
```
PeterOrneholm: Nice, but I would move it to the BankId Controller.So this will look like:

`public BankIdLoginApiErrorResponse(string errorMessage, bool retry)`

span: Of course. Glad someone is thinking!

PeterOrneholm:
Had a last look localy, and realizes that the error message briefly flashes in red between the QR codes.

Looking at it all again, I'm proposing a change like this:
image

The issue now is that the server actually returns a 500 and people will get a ton of 500 in their logs. As this is an expected "error" I would like to move the logic to the server.

By passing the attempts count to the server, wer don't need to store a state there. And we can return a "retryLogin" flag on the checkStatus itslelf.

Makes sense?

span:
Makes total sense. Did I get it right?

PeterOrneholm:
> Makes total sense. Did I get it right?

Almost, it fails on the client now. Update serverside to something like this and we should be fine:

```csharp
        private ActionResult CollectFailure(CollectResponse collectResponse, string statusMessage, int autoStartAttempts)
        {
            _logger.BankIdCollectFailure(collectResponse.OrderRef, collectResponse.GetCollectHintCode());

            var retryLogin = autoStartAttempts < MaxRetryLoginAttempts
                             && CollectHintCode.StartFailed.Equals(collectResponse.GetCollectHintCode());

            if (retryLogin)
            {
                return Ok(BankIdLoginApiStatusResponse.Pending(statusMessage, true));
            }
            else
            {
                return BadRequest(new BankIdLoginApiErrorResponse(statusMessage));
            }
        }
```

You need to add retryLogin to `BankIdLoginApiStatusResponse.Pending` and remove it from `BankIdLoginApiErrorResponse`.

span:
Ah!

I added a new response  `Retry` instead of using `Pending` to keep them separate. What do you think; :+1: or 👎

span:
Pushed a rebase from master.

PeterOrneholm:
> I added a new response `Retry` instead of using `Pending` to keep them separate. What do you think; 👍 or 👎

Nice! Will merge if the build succeeds.

PeterOrneholm:
Merged!

PR #163, opened 2019-11-17
Enable "C# 8 nullable reference types"

PeterOrneholm:
Fixes #114.

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.Api/BankIdApiException.cs
+++ b/src/ActiveLogin.Authentication.BankId.Api/BankIdApiException.cs
@@ -18,7 +18,7 @@ namespace ActiveLogin.Authentication.BankId.Api
         {
         }
 
-        internal BankIdApiException(ErrorCode errorCode, string errorDetails, Exception inner)
+        internal BankIdApiException(ErrorCode errorCode, string errorDetails, Exception? inner)
             : base($"{errorCode}: {errorDetails}", inner)
         {
             ErrorCode = errorCode;
--- a/src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
+++ b/src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
@@ -81,21 +81,32 @@ namespace ActiveLogin.Authentication.BankId.Api
 
         public async Task<AuthResponse> AuthAsync(AuthRequest request)
         {
-            var response = await GetOrderReponseAsync(request?.PersonalIdentityNumber, request?.EndUserIp).ConfigureAwait(false);
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            var response = await GetOrderResponseAsync(request.PersonalIdentityNumber, request.EndUserIp).ConfigureAwait(false);
             return new AuthResponse(response.OrderRef, response.AutoStartToken);
         }
 
         public async Task<SignResponse> SignAsync(SignRequest request)
         {
-            var response = await GetOrderReponseAsync(request?.PersonalIdentityNumber, request?.EndUserIp).ConfigureAwait(false);
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            var response = await GetOrderResponseAsync(request.PersonalIdentityNumber, request.EndUserIp).ConfigureAwait(false);
             return new SignResponse(response.OrderRef, response.AutoStartToken);
         }
 
-        private async Task<OrderResponse> GetOrderReponseAsync(string personalIdentityNumber, string endUserIp)
+        private async Task<OrderResponse> GetOrderResponseAsync(string? personalIdentityNumber, string endUserIp)
         {
             await SimulateResponseDelay().ConfigureAwait(false);
 
-            if (string.IsNullOrEmpty(personalIdentityNumber))
+            var personalIdentityNumberOrDefault = _personalIdentityNumber;
+            if (personalIdentityNumber == null || string.IsNullOrWhiteSpace(personalIdentityNumber))
             {
                 personalIdentityNumber = _personalIdentityNumber;
             }
@@ -153,7 +164,7 @@ namespace ActiveLogin.Authentication.BankId.Api
             return response;
         }
 
-        private CompletionData GetCompletionData(string endUserIp, CollectStatus status, string personalIdentityNumber)
+        private CompletionData? GetCompletionData(string endUserIp, CollectStatus status, string personalIdentityNumber)
         {
             if (status != CollectStatus.Complete)
             {
--- a/src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
+++ b/src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
@@ -144,23 +144,21 @@ namespace ActiveLogin.Authentication.BankId.Api
             var auth = _auths[request.OrderRef];
             var status = GetStatus(auth.CollectCalls);
             var hintCode = GetHintCode(auth.CollectCalls);
-            var completionData = GetCompletionData(auth.EndUserIp, status, auth.PersonalIdentityNumber);
 
-            var response = new CollectResponse(auth.OrderRef, status.ToString(), hintCode.ToString(), completionData);
-
-            if (status == CollectStatus.Complete)
-            {
-                if (_auths.ContainsKey(request.OrderRef))
-                {
-                    _auths.Remove(request.OrderRef);
-                }
-            }
-            else
+            if (status != CollectStatus.Complete)
             {
                 auth.CollectCalls += 1;
+                return new CollectResponse(auth.OrderRef, status.ToString(), hintCode.ToString());
             }
 
-            return response;
+            if (_auths.ContainsKey(request.OrderRef))
+            {
+                _auths.Remove(request.OrderRef);
+            }
+
+            var completionData = GetCompletionData(auth.EndUserIp, status, auth.PersonalIdentityNumber);
+
+            return new CollectResponse(auth.OrderRef, status.ToString(), hintCode.ToString(), completionData);
         }
 
         private CompletionData? GetCompletionData(string endUserIp, CollectStatus status, string personalIdentityNumber)
--- a/src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
+++ b/src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
@@ -105,7 +105,6 @@ namespace ActiveLogin.Authentication.BankId.Api
         {
             await SimulateResponseDelay().ConfigureAwait(false);
 
-            var personalIdentityNumberOrDefault = _personalIdentityNumber;
             if (personalIdentityNumber == null || string.IsNullOrWhiteSpace(personalIdentityNumber))
             {
                 personalIdentityNumber = _personalIdentityNumber;
--- a/src/ActiveLogin.Authentication.BankId.Api/Errors/BankIdApiErrorHandler.cs
+++ b/src/ActiveLogin.Authentication.BankId.Api/Errors/BankIdApiErrorHandler.cs
@@ -18,7 +18,7 @@ namespace ActiveLogin.Authentication.BankId.Api.Errors
             }
             catch (HttpRequestException e)
             {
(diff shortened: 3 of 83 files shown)


src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
```diff
@@ -81,21 +81,32 @@ public TimeSpan Delay
...
             return new SignResponse(response.OrderRef, response.AutoStartToken);
         }
 
-        private async Task<OrderResponse> GetOrderReponseAsync(string personalIdentityNumber, string endUserIp)
+        private async Task<OrderResponse> GetOrderResponseAsync(string? personalIdentityNumber, string endUserIp)
         {
             await SimulateResponseDelay().ConfigureAwait(false);
 
-            if (string.IsNullOrEmpty(personalIdentityNumber))
+            var personalIdentityNumberOrDefault = _personalIdentityNumber;
```
nikolaykrondev: Is there a hidden usage of this variable? I don't see it within this method.

src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
```diff
@@ -153,7 +164,7 @@ public async Task<CollectResponse> CollectAsync(CollectRequest request)
             return response;
         }
 
-        private CompletionData GetCompletionData(string endUserIp, CollectStatus status, string personalIdentityNumber)
+        private CompletionData? GetCompletionData(string endUserIp, CollectStatus status, string personalIdentityNumber)
         {
             if (status != CollectStatus.Complete)
```
nikolaykrondev: We do double check for CollectionStatus.Complete, if we could move call of GetCompletionData within true case of if statement at line 152 we don't need to check for status here and we can return not nullable CompletionData

src/ActiveLogin.Authentication.BankId.Api/Errors/BankIdApiErrorHandler.cs
```diff
@@ -27,7 +27,7 @@ public static async Task EnsureSuccessAsync(HttpResponseMessage httpResponseMess
             }
         }
 
-        private static async Task<Error> TryGetErrorAsync(HttpResponseMessage httpResponseMessage)
+        private static async Task<Error?> TryGetErrorAsync(HttpResponseMessage httpResponseMessage)
```
nikolaykrondev: I know we are in the nullable types PR :) but I would like to use as less as possible nullable occasions. Here we can keep it as not null Error, but within instead of returning null, we can return Error.Empty - no-opt solution.
Implementation inside of Error class could be:
`public static Error Empty => new Error(string.Empty, string.Empty);`
It is much better practice to compare later on.

nikolaykrondev: Line 21 here would be:
`if (error != Error.Empty)`
Line 45 here would be 
`return Error.Empty;`

src/ActiveLogin.Authentication.BankId.Api/Models/AuthRequest.cs
```diff
@@ -35,7 +35,7 @@ public AuthRequest(string endUserIp)
         /// The personal number of the user. 12 digits, century must be included (YYYYMMDDSSSC).
         /// If the personal number is excluded, the client must be started with the AutoStartToken returned in the response.
         /// </param>
-        public AuthRequest(string endUserIp, string personalIdentityNumber)
+        public AuthRequest(string endUserIp, string? personalIdentityNumber)
```
nikolaykrondev: Since we do not handle null case here, isn't it better to create additional constructor that does not take `personalIdentityNumber`?

src/ActiveLogin.Authentication.BankId.Api/Models/AuthRequest.cs
```diff
@@ -54,7 +54,7 @@ public AuthRequest(string endUserIp, string personalIdentityNumber)
         /// If the personal number is excluded, the client must be started with the AutoStartToken returned in the response.
         /// </param>
         /// <param name="requirement">Requirements on how the auth or sign order must be performed.</param>
-        public AuthRequest(string endUserIp, string personalIdentityNumber, Requirement requirement)
+        public AuthRequest(string endUserIp, string? personalIdentityNumber, Requirement requirement)
```
nikolaykrondev: ... then we don't need to mark this one as null either

src/ActiveLogin.Authentication.BankId.Api/BankIdApiException.cs
```diff
@@ -18,7 +18,7 @@ internal BankIdApiException(ErrorCode errorCode, string errorDetails)
         {
         }
 
-        internal BankIdApiException(ErrorCode errorCode, string errorDetails, Exception inner)
+        internal BankIdApiException(ErrorCode errorCode, string errorDetails, Exception? inner)
```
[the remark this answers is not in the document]
PeterOrneholm: I think both of the ones above point to this one, to not duplicate code. Wouldn't he have to do so otherwise?

src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
PeterOrneholm: Great catch, it's a leftover. Will remove.

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
```diff
@@ -92,8 +92,18 @@ public SignRequest(string endUserIp, string userVisibleData, byte[] userNonVisib
         /// If the personal number is excluded, the client must be started with the AutoStartToken returned in the response.
         /// </param>
         /// <param name="requirement">Requirements on how the auth or sign order must be performed.</param>
-        public SignRequest(string endUserIp, string userVisibleData, byte[] userNonVisibleData, string personalIdentityNumber, Requirement requirement)
+        public SignRequest(string endUserIp, string userVisibleData, byte[]? userNonVisibleData, string? personalIdentityNumber, Requirement? requirement)
```
nikolaykrondev: The same here - one more constructor instead of implicit null in the arguments?

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
```diff
@@ -75,9 +75,14 @@ public class BankIdApiController : Controller
...
         {
+            if (request.LoginOptions == null)
+            {
+                throw new ArgumentNullException(nameof(request.LoginOptions));
+            }
+
             var unprotectedLoginOptions = _loginOptionsProtector.Unprotect(request.LoginOptions);
 
-            SwedishPersonalIdentityNumber personalIdentityNumber;
+            SwedishPersonalIdentityNumber? personalIdentityNumber;
```
nikolaykrondev: I suppose that here we expect null from unprotectedLoginOptions?

src/ActiveLogin.Authentication.BankId.Api/BankIdSimulatedApiClient.cs
PeterOrneholm: Is this really possible? We need the CollectResponse even in the false case.

src/ActiveLogin.Authentication.BankId.Api/BankIdApiException.cs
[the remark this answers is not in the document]
nikolaykrondev: Hm... yes and no :) I just personally don't like implicit stuff.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiCancelRequest.cs
```diff
@@ -10,8 +10,8 @@ internal BankIdLoginApiCancelRequest()
         }
 
         [Required]
-        public string OrderRef { get; set; }
+        public string? OrderRef { get; set; }
```
nikolaykrondev: Why can this be null? First it is marked as required, second - what is the context to use cancel without the orderref? :)

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiInitializeRequest.cs
```diff
@@ -9,13 +9,12 @@ internal BankIdLoginApiInitializeRequest()
             
         }
 
-
-        public string PersonalIdentityNumber { get; set; }
+        public string? PersonalIdentityNumber { get; set; }
 
         [Required]
-        public string ReturnUrl { get; set; }
+        public string? ReturnUrl { get; set; }
```
nikolaykrondev: The same here, I just curios how runtime behaves with this combination

src/ActiveLogin.Authentication.BankId.Api/Errors/BankIdApiErrorHandler.cs
PeterOrneholm: Yeah, I like the approach, but don't we expose that "publicly" just to solve an internal thing in this class? From our consumers point of view, we would have an `.Empty` on this model class, but on no other.

src/ActiveLogin.Authentication.BankId.Api/Models/AuthRequest.cs
PeterOrneholm: Agree, there shouldn't be a null thing there, good catch!

src/ActiveLogin.Authentication.BankId.Api/Models/SignRequest.cs
PeterOrneholm: The recommendation is to only do additions on parameters, not change the order of them:
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading?redirectedfrom=MSDN

> X AVOID being inconsistent in the ordering of parameters in overloaded members. Parameters with the same name should appear in the same position in all overloads.

So, we need to have one with all the params where they can be nullable. We could have a few more to catch the secario where you have userNonVisibleData but not personalIdentityNumber, but I think it might be overkill for now? We end up with a ton of overloads :)

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdApiController.cs
PeterOrneholm: Yes, that can be null if not PIN is passed, which is the default case.

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiCancelRequest.cs
PeterOrneholm: We do require it, but as this is something passed from the client, it can be null. If the client don't set a value for it, it will be null. The required flag only ells validation it is required. I know it looks strange, but the nullable stuff is more "can this ever be in a state where it is set to null", and it can be in this case :)

src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Models/BankIdLoginApiInitializeRequest.cs
PeterOrneholm: Same thing there, ASP.NET Core will (from my understanding) set it to null if no valye is passed from the client.

PeterOrneholm: See details here:
https://github.com/aspnet/AspNetCore/issues/14451#issuecomment-535403894

> As far as I know, modelbinding is not affected in any way by the use of Nullable types in C#. Nullable types help the compiler specify the behavior, but don't offer any runtime guarantee about it nor extend the range of values a type can take.

Basicallt, C#8 nullable are pure compile time helper for the developer. A non nullable string can still be set to null... :)

PR #179, opened 2019-12-03
Added AngleSharp for querying Html in tests

### LINKED ISSUE #147: Use AngleSharp in tests
From @PeterOrneholm in #138 

>  I recently stumbled upon AngleSharp for parsing HTML in C#. Microsoft seems to recommend it here: https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-3.0
>
> If I get it right, you can parse the DOM using the lib, query for the element you like and check a specific prop. That would make our tests less fragile.

This is a great idea for the tests and will make it much easier to test the different paths in the authentication flow.

martinsjolander:
This PR fixes #147 .


PeterOrneholm:
Really nice work @martinsjolander!

PR #209, opened 2020-01-06
Change secret key to secret name for azure certificates

### LINKED ISSUE #208: Azure Key Vault secret api uses key instead of name
**Describe the bug**
In the `ClientCertificateFromAzureKeyVaultOptions` the "secret name" is defined as the "secret key" which can be confusing.

BankID Azure

**NuGet package version**
Latest master

**Additional context**
The Microsoft API uses "name" and not "key" when fetching a secret from KeyVault. We should do the same.

https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.secrets.secretclient.getsecret?view=azure-dotnet#Azure_Security_KeyVault_Secrets_SecretClient_GetSecret_System_String_System_String_System_Threading_CancellationToken_

span:
This PR will clarify the naming of the certificate/secret name in
the options for using client certificates in Azure.

This PR fixes #208 .

### DIFF
--- a/docs/getting-started-bankid.md
+++ b/docs/getting-started-bankid.md
@@ -41,16 +41,16 @@ These are only necessary if you plan to store your certificates in Azure KeyVaul
         "AzureAdClientSecret": "",
 
         "AzureKeyVaultUri": "TODO-ADD-YOUR-VALUE",
-        "AzureKeyVaultSecretKey": "TODO-ADD-YOUR-VALUE"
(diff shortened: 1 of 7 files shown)


PeterOrneholm:
Thanks for this fix!

I've updated the release pipeline for our sample to use the new name.
I'll merge once the tests are green.

span:
Not sure what is going on here. I'll dig around.

PR #228, opened 2020-02-27
Allow overriding retreival of EndUserIp

### LINKED ISSUE #221: Allow overriding retreival of EndUserIp
We have had multiple issues (#152, #213) related to what we use as a value for EndUserIp. Seems that our current solution `HttpContext.Connection.RemoteIpAddress.ToString()` might be too naive in scenarios with Docker, Proxies etc.

- ActiveLogin.Authentication.BankId.AspNetCore

Add an interface that has one method that should return the value to use for EndUserIp. The default implementation implemented in #215 but can be overridden depending on the scenario. We should also add extension methods to make that easier.

Suggestion:

An implementation could take whatever dependency it needs, but as HttpContext will be used in 99.9% of the implementations, that's what we will send to the method.

```csharp
public interface IBankIdEndUserIpResolver {
      string GetEndUserIp(HttpContext context);
}

public static class BankIdBuilderExtensions
    {
        public static IBankIdBuilder UseEndUserIpResolver<TEndUserIpResolver>(this IBankIdBuilder builder) where TEndUserIpResolver : IBankIdEndUserIpResolver
        {
            ...
        }

        public static IBankIdBuilder UseEndUserIpResolver(this IBankIdBuilder builder, Func<HttpContext, string> configureEndUserIpResolver)
        {
            ...
        }
    }
```

PeterOrneholm:
This PR fixes #221.

- [x] Make end user IP resolver and interface
- [x] Add extension methods for easier usage
- [x] Documentation

### DIFF
--- a/docs/getting-started-bankid.md
+++ b/docs/getting-started-bankid.md
@@ -360,6 +360,62 @@ The default implementation will log all data to the tracelog. If you want to rem
 services.RemoveAll(typeof(IBankIdResultStore));
 ```
 
+### Resolve the end user ip
+
+In some scenarios, like running behind a proxy, you might want to resolve the end user IP yourself and override the default implementaion.
+
+Either register a class implementing `IEndUserIpResolver`:
+
+```csharp
+builder.UseEndUserIpResolver<EndUserIpResolver>();
+```
+
+Or use the shorthand version:
+
+```csharp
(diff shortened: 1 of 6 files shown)


PeterOrneholm:
Would you like to have a quick look at this on @span and @magohl?
Should make it easier to resolve the end user IP yourself in special scenarios.

PR #250, opened 2020-05-15
Change redirect scheme for android (242)

### LINKED ISSUE #242: Change redirect scheme for Android >= 6 as per new guidelines
Last week BankID released a new version (v3.2.3) of their [BankID Relying Party Guidelines](https://www.bankid.com/assets/bankid/rp/bankid-relying-party-guidelines-v3.2.4.pdf). The change is that it's now recommended to redirect Android devices that runs at least Android 6 to the https://app.bankid.com/ domain instead of bankid:/// scheme. We already do the same for iOS.

- BankID

Update `BankIdSupportedDevice.cs` to allow for setting an (optional) OS version. Let the default implementation (`BankIdSupportedDeviceDetector.cs`) detect that for Android and then use the OS version in `BankIdLauncher.cs` to select the correct prefix.

**Additional context**
Might be worth checking out if we can automatically launch BankId on Android as well. Right now we show the button on all versions of Android and the user has to click it.

elinohlsson:
Fixes #242.

Detect device Os version for Android and change redirect scheme for Android version >= 6 as per new guidelines.


PeterOrneholm:
Nice work! Will make the experience much better on modern Android devices.

PR #263, opened 2020-05-27
Add package with device detection using ua parser (256)

### LINKED ISSUE #256: Add package with device detection using ua_parser
In Active Login we do quite a lot of device/browser detection instead of feature detection. This is in line with the BankID Relaying party Guidelines and is, in most cases, because we can't feature detect these things. For example, we might redirect to a URL to launch the BankID app and to know what URL we should use we need to actually launch the application - and these kind of things is hard/impossible to do feature detection for. Most of this logic is encapsulated into `IBankIdLauncher` and `IBankIdSupportedDeviceDetector`.

One (of many) issues with browser detection is that [user agents are a mess](https://webaim.org/blog/user-agent-string-history/) and are not that very easy to map into browsers, because they all say they are everything. In Active Login we have an implementation that supports the ~top 5 most common browsers on both iOS and Android, but that's what we have limited our implementation to.

The issue here is that we get a few false positives detecting Chrome as the browser. `IBankIdLauncher` will check if the browser is Chrome, and then use the new launch URL (#242). This feature only works in Chrome and will cause the flow to be broken in other browsers. But because basically every browser on Android identifies as Chrome in their UA, we need to check that it's not any of the other known ones.

We feel that the top 5 is good enough for our default implementation, but to get better support for other browsers, a better user agent parsing could be added.

- BankID

Add another package (ActiveLogin.Authentication.BankId.BankId.AspNetCore.UAParser) that implements `IBankIdSupportedDeviceDetector` using [uap-csharp](https://github.com/ua-parser/uap-csharp). Then add an extension method (`.UseUaParserDeviceDetection()`) for easy usage.

Optimally we would be less dependent on the actual browser/device but so far we have not found any way around this for the features we are suing it for today.

elinohlsson:
This PR fixes #256 .

Implementation of IBankIdSupportedDeviceDetector based on [uap-csharp](https://github.com/ua-parser/uap-csharp) for better detection of third party browsers.


src/ActiveLogin.Authentication.BankId.AspNetCore.UAParser/UAParserDeviceDetector.cs
```diff
@@ -0,0 +1,156 @@
+using System;
+using ActiveLogin.Authentication.BankId.AspNetCore.SupportedDevice;
+using UAParser;
+
+namespace ActiveLogin.Authentication.BankId.AspNetCore.UAParser
+{
+    public class UAParserDeviceDetector : IBankIdSupportedDeviceDetector
```
PeterOrneholm, line 7: As this is a public class, we can add a small XML comment saying what it does.

src/ActiveLogin.Authentication.BankId.AspNetCore.UAParser/UAParserDeviceDetector.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        private bool IsMobileBrowser(UserAgent userAgent)
+        {
+            return IsBrowser(userAgent, "mobile");
+        }
+
+        private bool IsBrowser(UserAgent userAgent, string browser)
+        {
+            return userAgent.Family.ToLower().Contains(browser.ToLower(), StringComparison.InvariantCulture);
```
PeterOrneholm, line 123: We've done containts before, but maybe we can be more exact in our match now? Not sure in what format UA parser returns it. Could be do (for example) `StartsWith`?

Also, I think we could do `StringComparer.InvariantCultureIgnoreCase` and skip `ToLower()`.

The same applies to `IsOs()` below.

src/ActiveLogin.Authentication.BankId.AspNetCore.UAParser/UAParserDeviceDetector.cs
elinohlsson: Added XML comment.

src/ActiveLogin.Authentication.BankId.AspNetCore.UAParser/UAParserDeviceDetector.cs
elinohlsson: Replaced ToLower() with StringComparer.InvariantCultureIgnoreCase.

elinohlsson:
Updated readme file and changed the sample projects to use the new uaparser.

src/ActiveLogin.Authentication.BankId.AspNetCore.UAParser/UAParserDeviceDetector.cs
PeterOrneholm: Nice! Did you see if there was any way to narrow it down to (something like StartsWith) instead of Contains?

elinohlsson: I have noticed that the string in ClientInfo.UA.Family could be e.g. "Mobile Safari" or "Chrome Mobile" i.e. it does not always start with the name of the browser.

Issue #248, opened 2020-05-15
Update Active Login to .NET 6


PeterOrneholm:
**Is your feature request related to a problem? Please describe.**
.NET 6 (LTS) is released. We should upgrade to that version. Most specifically ASP.NET on .NET 6. This to ensure we won't have to much work when wee need to upgrade.

**What area is it related to**
- BankID
- GrandID

**Additional context**
Should we consider dropping support for .NET-standard?

PR #279, opened 2020-09-28
Made model constructors public to enable deserialization

ChrisKlug:
In .NET 5, it seems as though the JSON deserialization currently requires public constructors for it to work. Because of this, I have updated the constructors for the request models to make sure it works.

- [x] Updated request models to have public constructors


nikolaykrondev:
Is it also working with private setters for the fields now?

ChrisKlug:
Right now all of them seem to have public setters, which I think is an ok choice to be honest. I believe the serializer supports private setters, but I don't see a real need for it in this case

PeterOrneholm:
Looks good. Congratulations to your first PR for Active Login :)

Adding #248 as it is related.

PR #292, opened 2021-06-14
Update FAQ with certificate handling when running in Linux

### LINKED ISSUE #290: Improved Linux documentation
**Certificate problem in Linux**
I have spend some time with certificate problems when running in Linux (Docker). My initial solution was to embedd the BankId root certificates as files as Embedded resources and then read that resource. I made an byte array of the file and used that byte array as argument for X509Certificate2. Something like this:

```csharp
var assembly = typeof(AuthBuilderExtensions).Assembly;
var resourceName = $"MyNamespace.BankIdRootCertificate-{environment}.crt";

using var certStream = assembly.GetManifestResourceStream(resourceName);
using var memoryStream = new MemoryStream((int)certStream.Length);

certStream.CopyTo(memoryStream);
var bytes = memoryStream.ToArray();
return new X509Certificate2(bytes, string.Empty, X509KeyStorageFlags.MachineKeySet);
```

This works perfectly on Windows, but fails with the following Exception in Linux:

```
System.Security.Cryptography.CryptographicException: ASN1 corrupted data.
 ---> System.Formats.Asn1.AsnContentException: The encoded length exceeds the maximum supported by this library (Int32.MaxValue).
   at System.Formats.Asn1.AsnDecoder.ReadLength(ReadOnlySpan`1 source, AsnEncodingRules ruleSet, Int32& bytesConsumed)
   at System.Formats.Asn1.AsnDecoder.ReadEncodedValue(ReadOnlySpan`1 source, AsnEncodingRules ruleSet, Int32& contentOffset, Int32& contentLength, Int32& bytesConsumed)
   at Internal.Cryptography.Pal.UnixPkcs12Reader.ParsePkcs12(ReadOnlySpan`1 data)
   --- End of inner exception stack trace ---
   at Internal.Cryptography.Pal.StorePal.FromBlob(ReadOnlySpan`1 rawData, SafePasswordHandle password, X509KeyStorageFlags keyStorageFlags)
   at System.Security.Cryptography.X509Certificates.X509Certificate2Collection.Import(ReadOnlySpan`1 rawData, ReadOnlySpan`1 password, X509KeyStorageFlags keyStorageFlags)
   at System.Security.Cryptography.X509Certificates.X509Certificate2Collection.Import(Byte[] rawData, String password, X509KeyStorageFlags keyStorageFlags)
```

It took me a while before I resolved that this was the problem and I thought I could at least save other some time.

**My solution**
Instead of embedding the file as a resource, I instead stored the content between Begin certificate and End certificate in a resource file. This data is base64 encoded and on Linux it must be decoded. My code now looks like this:

```
var rootCertEncoded = CertificateResource.BankIdRootTest;
var rootCertBytes = Convert.FromBase64String(rootCertEncoded);

return new X509Certificate2(rootCertBytes, (System.Security.SecureString)null, X509KeyStorageFlags.MachineKeySet);
```

Less code and simpel to read.

BankId documentation.

Maybe this could be added to the Readme or at least exists here in the issue tracker if someone else gets this problem.

equist:
This PR fixes #290.

### DIFF
--- a/docs/getting-started-bankid.md
+++ b/docs/getting-started-bankid.md
@@ -574,3 +574,18 @@ The UI is bundled into the package as a Razor Class Library, a technique that al
 ### Can the messages be localized?
 
 The messages are already localized to English and Swedish using the official recommended texts. To select what texts that are used you can for example use the [localization middleware in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-2.1#localization-middleware).
+
+### How can the certificates be handled when running in Linux?
+
+X509Certificate2 can not be handled in the same way when running in Linux as on Windows. The certificate is Base64 encoded and must be decoded before creating the X509Certificate2 instance. Below is an example for the BankId root certificate:
+
+Copy the content between Begin certificate and End certificate, and paste it into a resource string in a Resource.resx file.
+
+With the certificate in the reosurce, this code can be used to create the X509Certificate2 instance. Note the second line that decodes the Base64 string.
+
+```csharp
+var rootCertEncoded = CertificateResources.BankIdRootTestCertificate;
+var rootCertBytes = Convert.FromBase64String(rootCertEncoded);
+
+return new X509Certificate2(rootCertBytes, string.Empty, X509KeyStorageFlags.MachineKeySet);
+```


PeterOrneholm:
Great, thanks!

PR #302, opened 2021-12-22
Add response for animated Qr tokens

### LINKED ISSUE #275: Add support for API V5.1
BankID API V5.1 was recently released, adding support for animated QR codes. We should add the extra field(s) in our API wrapper.

BankID

Add/parse the new fields:
- `qrStartToken`
- `qrStartSecret`

**Additional context**
Described in [BankID Relying Party Guidelines v3.4](https://www.bankid.com/assets/bankid/rp/bankid-relying-party-guidelines-v3.4.pdf) under *14.1.3 Response from auth and sign*.

ellinge:
This PR fixes #275 .

Added a simplified fallback on autoStartTokenRequired for backward compatiblitity (string.replace). Thought about adding another class for v5 regarding requirement or changing it on serialization but felt a bit overkill. One perhaps could add both datamembers as well.

I've tried it out with generation of qr token with:
```csharp
        public AuthResponse Response { get; }

        public DateTime InitTime { get; } = DateTime.Now;

        /// <summary>
        /// Calculates QR token according to https://www.bankid.com/utvecklare/guider/teknisk-integrationsguide/qrkoder
        /// </summary>
        /// <param name="orderref"></param>
        /// <returns></returns>
        public string GetQrToken(string orderref)
        {
            if(Response?.OrderRef is string refOrderRef &&
                string.Equals(refOrderRef, orderref, StringComparison.InvariantCultureIgnoreCase) &&
                Response.QrStartSecret is string startSecret && Response.QrStartToken is string startToken)
            {
                var time = ((int)(DateTime.Now - InitTime).TotalSeconds).ToString();

                using (var mac = new HMACSHA256(Encoding.ASCII.GetBytes(startSecret)))
                {
                    var qrAuthBytes = mac.ComputeHash(Encoding.ASCII.GetBytes(time));
                    var qrAuthCode = string.Concat(qrAuthBytes.Select(b => string.Format("{0:x2}", b)));
                    return string.Join(".", new[] { "bankid", startToken, time, qrAuthCode });
                }
            }
            return null;
        }
```

Which works fine.

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.Api/BankIdHttpClientExtensions.cs
+++ b/src/ActiveLogin.Authentication.BankId.Api/BankIdHttpClientExtensions.cs
@@ -3,6 +3,7 @@ using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
 using ActiveLogin.Authentication.BankId.Api.Errors;
+using ActiveLogin.Authentication.BankId.Api.Models;
 using ActiveLogin.Authentication.Common.Serialization;
 
 namespace ActiveLogin.Authentication.BankId.Api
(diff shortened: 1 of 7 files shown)


PeterOrneholm:
Thank you for this PR! Will try to find time during the week to review it.

PeterOrneholm:
Thank you for this great work @ellinge!

I've changed the base branch of your PR to a feature branch in our repo. There are a few more things related to v5.1 I'd like to add when doing this. So will continue your work from there.

Also, this release will only support 5.1 of the API - so I will remove your backward compatibility check. If support is needed for 5.0 people can always stay on their old version on NuGet.

PeterOrneholm:
@ellinge I'm happy to say I've just released the code in your PR to NuGet :) Thank you!
https://www.nuget.org/packages/ActiveLogin.Authentication.BankId.Api/5.0.0-alpha-2

All packages for 5.0.0 are flagged as alpha, but I'm pretty confident that the .Api package will stay pretty much solid until stable release.

PR #309, opened 2022-01-17
Implement invalid state handling

### LINKED ISSUE #298: Invalid state cookie
**Describe the bug**
Transient errors that say "Invalid state cookie" from the `BankIdHandler`.

This report is from Slack user Anton K:
I can provoke this error myself by removing the “__ActiveLogin.BankIdState”-cookie or by modifying the value within the cookie during the login flow. This got me thinking that some devices/browser may have problems setting the cookie. The cookie is set as HTTP-Only, Secure and Lax. 

BankId

**To Reproduce**
1. Remove “__ActiveLogin.BankIdState” cookie
2. See error

**Expected behavior**
Successful signin.

**Screenshots**
```
[2021-11-29 09:57:57 Error] Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware
An unhandled exception has occurred while executing the request.
System.Exception: An error was encountered while handling the remote login.
 ---> System.Exception: Invalid state cookie
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.HandleRequestAsync()
   at IdentityServer4.Hosting.FederatedSignOut.AuthenticationRequestHandlerWrapper.HandleRequestAsync()
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at IdentityServer4.Hosting.BaseUrlMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware.<Invoke>g__Awaited|6_0(ExceptionHandlerMiddleware middleware, HttpContext context, Task task)
```

**NuGet package version**
4.0.0

**Runtime version**
netcoreapp3.1

 - ios
 - safari

Only seen on ios and safari. Unknown versions.

User reports being able to sign in "sometimes" and then not so it seems to be a transient error where we sometimes get into a bad state. Perhaps due to refresh/reload/back button pressing or using old/stale data.

### LINKED ISSUE #301: Allow for custom handling when the state cookie is missing
As described in #298 we have found scenarios where the state cookie might be missing. Adding #300 would also introduce scenarios where state might be available but have expired.

We should handle that scenario in a way that it has some default implementation but is customizable for those who want/needs that.

BankID

Check in `BankIdController` > `Login` if the state cookie is still there and is valid. If not, let the injected handler handle it.

```csharp
if (IsInvalidStateCookie())
{
  if(unprotectedLoginOptions == null) {
    await _bankIdInvalidStateHandler.HandleAsync(context);
  } else {
   await _bankIdInvalidStateHandler.HandleAsync(context, unprotectedLoginOptions.CancelReturnUrl);
  }

    return;
}
```

A default implementation could use the cancelUrl, like this:

```csharp
public BankIdCancelUrlInvalidStateHandler : IBankIdInvalidStateHandler {
  public TaskHandleAsync(HttpContext context) {
    context.Response.Redirect("/");
  }

  public TaskHandleAsync(HttpContext context, string cancelReturnUrl) {
    context.Response.Redirect(cancelReturnUrl);
  }
}
```

PeterOrneholm:
This PR fixes #301 and fixes #298.

- [x] Add handling when the state is invalid
- [x] Add default implementation that uses cancel url when state is invalid
- [x] Add integration test

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdController.cs
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdController.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Threading.Tasks;
 using ActiveLogin.Authentication.BankId.Api.UserMessage;
 using ActiveLogin.Authentication.BankId.AspNetCore.Areas.BankIdAuthentication.Models;
 using ActiveLogin.Authentication.BankId.AspNetCore.DataProtection;
 using ActiveLogin.Authentication.BankId.AspNetCore.Models;
+using ActiveLogin.Authentication.BankId.AspNetCore.StateHandling;
 using ActiveLogin.Authentication.BankId.AspNetCore.UserMessage;
 using ActiveLogin.Authentication.Common.Serialization;
 using Microsoft.AspNetCore.Antiforgery;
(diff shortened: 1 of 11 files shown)


src/ActiveLogin.Authentication.BankId.AspNetCore/Areas/BankIdAuthentication/Controllers/BankIdController.cs
```diff
@@ -21,34 +23,55 @@ public class BankIdController : Controller
...
             if (!Url.IsLocalUrl(returnUrl))
             {
                 throw new Exception(BankIdConstants.InvalidReturnUrlErrorMessage);
             }
 
             var unprotectedLoginOptions = _loginOptionsProtector.Unprotect(loginOptions);
-            var antiforgeryTokens = _antiforgery.GetAndStoreTokens(HttpContext);
+            if (!HasStateCookie(unprotectedLoginOptions))
+            {
+                var invalidStateContext = new BankIdInvalidStateContext(unprotectedLoginOptions.CancelReturnUrl);
```
span, line 53: CHECK: Any reason we do not want to pass in all the options here?

PeterOrneholm: Yes - we gave it a thought and only want to expose as little as possible. The very moment we expose a thing, we need to maintain it for some time. In this case I only want to provide data that I think makes sense, which in this case is CancelReturnUrl. If we have a usecase for another of tjose properties I'm willing to add that to the "context" as well :)

PR #340, opened 2022-02-17
Support for user data on auth request

### LINKED ISSUE #328: Add support for userVisibleData, userNonVisibleData, userVisibleDataFormat on auth
Turns out BankID just released support for supplying data (visible and non visible) for auth. This had previously only been available for sign. We should add this to both the API wrapper but also making it possible to supply custom data per auth.

These fields have been added:
- `userVisibleData`
- `userNonVisibleData`
- `userVisibleDataFormat`

API, UI

For the API, something very similar to how auth works.

For the AspNet package I'm thinking some kind of interface you can plug into and given some context (not sure what is available right now, but HttpContext should be one of them) the function would be responsible for responding with that data.

The AuthRequest and SignRequest now seems to have exactly the same properties. I'd suggest making sure their constructors uses the same order for parameters. Maybe use a shared base class, not sure.

joanro123:
This PR fixes #328 .

- [x] Added userVisibleData, userNonVisibleData and userVisibleDataFormat to Auth request
- [x] Possible to add static data on startup
- [x] Added tests to test functionality

- [x] Code written
- [x] Test added

### DIFF
--- a/docs/bankid.md
+++ b/docs/bankid.md
@@ -850,6 +850,7 @@ builder.UseAuthRequestUserData(authUserData =>
     authUserData.UserVisibleData = message.ToString();
     authUserData.UserVisibleDataFormat = BankIdUserVisibleDataFormats.SimpleMarkdownV1;
 });
+```
 
 For more advanced scenarios, you can generate the user data dynamically by implementing `IBankIdAuthRequestUserDataResolver`:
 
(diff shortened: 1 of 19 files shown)


PeterOrneholm (approved):
Great work. Fixed a few things. Think we are ready to merge now

PR #350, opened 2022-03-03
Implement custom claims transformation

### LINKED ISSUE #337: Get Orderref as a txn claim
If the application that uses ActiveLogin BankId needs to keep an audit trail of the sign-in, it is not possible toady. We can store the completionData and orderRef and correlate it with a user, but it is not possible to correlate it to the exact sign-in. This means if the user sign-in, sign-out and then sign-in again we'll have two orderRef and completionData for the user, but it's not easy to correlate which orderRef corrsponds to which sign-in. Authentication Time is not good enough. Instead it would be nice if it was possible to get the orderRef as a claim on the user. The txn claim could preferably be used for this.

From OpenId Connect for Identity Assurance (https://openid.net/specs/openid-connect-4-identity-assurance-1_0.html):
The txn Claim as defined in [RFC8417] is used in the context of this extension to build audit trails across the parties involved in an OpenID Connect transaction.

IANA (https://www.iana.org/assignments/jwt/jwt.xhtml) list txn as a transaction identifier.

UI and issued user claims.

A txn claim with the orderRef value should be issued when the user is authenticated.

If not always issuing the claim, make a setting to enable the issuance of the txn claim.

### LINKED ISSUE #351: Issue correct exp claim
**Describe the bug**
The exp claim is issued as a date, it should be:

> Expiration time on or after which the ID Token MUST NOT be accepted for processing. The processing of this parameter requires that the current date/time MUST be before the expiration date/time listed in the value. Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time. See [RFC 3339](https://openid.net/specs/openid-connect-core-1_0.html#RFC3339) [RFC3339] for details regarding date/times in general and UTC in particular.

image

Claims issuing

** Note: We found this while refactoring the claims issuing part and will fix it as part of that PR.

### LINKED ISSUE #352: Remove built in support for Birthdate and Gender claim
Since the start, we've allowed (even if opt in) for issuing the claim _Birthdate_ and _Gender_ claim from the hints parsed by ActiveLogin.Identity. As these should be used with care, and used with the knowledge that these might not be 100% accurate - I'm suggesting we remove these as built in claims that can be issued.

Claims issuing

Remove them as optional claims and instead add documentation on how to add them as a custom claims transform,

PeterOrneholm:
This PR fixes #337 and fixes #351  and fixes #352.

- [X] Add support for customizing the claims issuing part
- [x] Fix the JWT exp bug #351
- [x] Add docs on how to issue txn claim #337 

- [X] Code written
- [x] Test added


joanro123 (approved):
Looking good!

PR #382, opened 2022-09-08
Replace bundled test certificate with unexpired cert

### LINKED ISSUE #381: Expired test certificate
The bundled test certificate https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/main/src/ActiveLogin.Authentication.BankId.Core/BankIdResources/FPTestcert3_20200618.p12 has expired.

from a series of magic openssl incantations:

```
$ openssl pkcs12 -in /mnt/c/Users/mikae/Downloads/FPTestcert3_20200618.p12 -passin pass:qwerty123 -passout pass:qwerty123 -out fp3.pem -nodes
$ openssl x509 -noout -enddate -in fp3.pem
notAfter=Sep  5 21:59:59 2022 GMT
```

mauve:
This PR fixes #381


PeterOrneholm:
Thanks @mauve! Looks like BankID issued a new one with really short notice and did not notify about it in any email or so.
I've pushed a small unit test to your PR that will fail the test if our bundled  certificates gets within 14 days of expiry date (as this seems to have been the time between the new cert being issued and the old one expiering). I added an GH Action to run the tests on schedule every day to ensure we will be notifed even if we don't push any code during that time.

mauve:
awesome @PeterOrneholm , I wonder if it makes sense to handle that error case in code somehow so that one would get an exception which says "Cert expired" instead of whatever it said :), took me waay too long to figure out that the cert was expired (and that we forgot to set `useBankIdCertificate: false` in our call to `UseTest...()`)

PeterOrneholm:
It really do. I think we have a few of those "internal exceptions" that we can foresee and throuw way better errors. I'll add an issue on this. 6.0.2 shoudl be out on nuget later today.

PR #408, opened 2023-04-26
Fix for AzureKeyVaultCertificateClient on Macos, 

### LINKED ISSUE #407: AzureKeyVaultCertificateClient fails on Macos
**Describe the bug**
When developing on Macos, `IBankIdBuilder.UseClientCertificateFromAzureKeyVault` fails with `MAC verification failed during PKCS12 import (wrong password?)` because Macos doesn't accept an empty string as a password.
However, passing `null` instead of `string.empty` in https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/main/src/ActiveLogin.Authentication.BankId.AzureKeyVault/AzureKeyVaultCertificateClient.cs#L72 resolves the problem.

API

**To Reproduce**
Use the API on Macos and try to start a signing with BankId using a certificate from Azure Key Vault.

**Expected behavior**
The signing process should start.

**NuGet package version**
5.0.0

**Runtime version**
Dotnet Core 6.0.408

**Desktop (please complete the following information):**
 - OS: Macos Ventura 13.3.1
 - Browser Chrome, Firefox

N/A

**Additional context**
The fix has been tested on Macos and Windows.

erikbarke:
This PR fixes #407.

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AzureKeyVault/AzureKeyVaultCertificateClient.cs
+++ b/src/ActiveLogin.Authentication.BankId.AzureKeyVault/AzureKeyVaultCertificateClient.cs
@@ -69,7 +69,7 @@ internal class AzureKeyVaultCertificateClient
     private X509Certificate2 GetX509Certificate2(byte[] certificate)
     {
         var exportedCertCollection = new X509Certificate2Collection();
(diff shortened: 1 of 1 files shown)


PeterOrneholm:
Thanks!

PR #437, opened 2024-01-08
Add client side retry logic

### LINKED ISSUE #433: Flow fails on third party browsers (Chrome/Edge) on iOS when returned from the BankID app
**Describe the bug**
We have seen (and been able to reproduce) that when using Chrome/Edge on iOS the flow works fine, but when returned to the browser it (sometimes) displays the error "Load failed". This seems to be a local error on iOS that happens when fetch() fails. Our idea is that this is related to that the browser tries to check the status even though the browser have been put into the "background" and is then prohibited to make a fetch call.

API, UI etc.

**To Reproduce**
1. Choose to sign in
2. Open BankID app
3. Put away phone and wait a while
3. Sign in
4. Go back to Edge/Chrome
5. Error: Load failed

**NuGet package version**
7.0.0

**Additional context**
Suggested solution:
* Add retry logic to the fetch calls
* Check if page is in focus (maybe usinghttps://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)

PeterOrneholm:
This PR fixes #433.

Adds 3 retries (with 1000ms delay inbetween) for `initialize`, `checkStatus` and `refreshQrCode`.

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Client/activelogin-main.ts
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Client/activelogin-main.ts
@@ -22,7 +22,8 @@ interface IBankIdUiScriptInitState {
 }
 
 function activeloginInit(configuration: IBankIdUiScriptConfiguration, initState: IBankIdUiScriptInitState) {
-    const fetchRetryRetryDelayMs: number = 1000;
+    const fetchRetryCountDefault: number = 3;
+    const fetchRetryDelayMs: number = 1000;
 
     // Pre check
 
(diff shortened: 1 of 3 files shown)


src/ActiveLogin.Authentication.BankId.AspNetCore/Client/activelogin-main.ts
```diff
@@ -126,7 +128,7 @@ function activeloginInit(configuration: IBankIdUiScriptConfiguration, initState:
             {
                 "returnUrl": returnUrl,
                 "uiOptions": protectedUiOptions
-            })
+            }, 3)
```
jakobehn, line 131: The 3 could be moved into a const, like the delayMs setting

jakobehn (approved):
LGTM

PR #445, opened 2024-02-22
Add the possibility to pass personal number requirement to sign using BankIdFlowService

nahojd:
This PR adds  the possibility to pass a personal number requirement to the /sign operation when calling it via the `BankIdFlowService`.

This is done by adding a property `RequirePersonalNumber` to `BankIdFlowOptions`, which by default will not be set, and then using that to set the personal number requirement in the `SignRequest`.


PeterOrneholm:
Thanks for the PR! Would like to se that you use `PersonalIdentityNumber` as type instead of string, to ensure a valid number is passed.

PeterOrneholm:
Looks good, down to the nitty gritty details now :) I would vote for calling it `RequiredPersonalIdentityNumber` instead. `Required` to show that it's not a bool (in contrast to the other properties) and `PersonalIdentityNumber` to align with our other naming conventions.

nahojd:
I called it PersonalNumber since that's what it's called in the BankId Requirement according to their documentation, but no problems, I have changed it to RequiredPersonalIdentityNumber.

PeterOrneholm:
Thanks for this! Will see when I can get this out, would probably be an 8.1 release.

Issue #479, opened 2024-12-20
Add support for setting the returnUrl as part of the request body in the auth and sign request instead of as part of the autolaunch url.


elinohlsson:
Support setting returnUrl in the request body of the auth and sign request. Currently it is only supported to provide it as part of the autolaunch url.

See information below from BankID regarding the return url in the auth/sign request.

Orders started on the same device as where the user's BankID is stored (started with autostart token) will call this URL when the order is completed.

Any return URL provided in the start URL when the BankID app was launched will be ignored. If the user has a version of the BankID app that does not support getting the returnUrl from the server, the order will be cancelled and the user will be asked to update their app.

The return URL you provide should include a nonce to the session. When the user returns to your app or web page, your service should verify that the order was completed successfully and that the device receiving the returnUrl is the same device that started the order.

Using this parameter will make your service more secure and strengthen the channel binding between you and the user.

Ensure that the cookie or user IP address has not changed from the starting page to the returnUrl page.

torselden:
When I read the documentation - again - from BankID, I interpret this more as "adding" support rather than "moving". Having the `returnUrl` in both the start url, and the request body for `/auth` and `/sign`, with the latter two taking precedence. 

We discussed this and we decided to initially leave the implementation of the `nonce` to the client.

Zonnex:
We should definitely start by adding support (seeing as this takes precedence), and be prepared to remove old way if needed. If possible, it's ideal if our api promotes the new way rather than the old way (to reduce possibility of confusion).

elinohlsson:
✅ This feature is now available in the stable release **v11.1.0**.  
See the [release notes](https://github.com/ActiveLogin/ActiveLogin.Authentication/releases/tag/v11.1.0) for details.  

I will now close this issue.

PR #499, opened 2025-03-26
feat(#479): set returnUrl in requests

Zonnex:
This PR fixes first part of #479, to assign the returnUrl for the AuthRequest/SignRequest we send to BankId. This URL takes precedence over the returnUrl given in the AutoStart functionality.

### DIFF
--- a/BREAKINGCHANGES.md
+++ b/BREAKINGCHANGES.md
@@ -25,14 +25,41 @@ ___Note:___ We might, and will probably, miss to document some of this - if so -
 Breaking changes between version 11.0.0 and 10.0.0
 
 ### Remove allowed risk Level
-Removed support for blocking auth and sign transactions based on risk level, since no longer supported by BankID. Use return risk instead and handle risk level in your application.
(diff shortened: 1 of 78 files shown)


elinohlsson (approved):
Looks good!

Liteolika (approved):
The security of the returnUrl feature relies entirely on correct implementation by library users. The potential main risk is improper validation of returnUrl and nonce by downstream developers which may expose open redirect or replay vulnerabilities.

A automated safeguard and/or validation of the returnUrl could be a future feature.

elinohlsson:
@Liteolika You're correct that this is a potential concern. In many scenarios (depending on device and OS), Active Login deliberately clears the returnUrl and replaces it with an empty string. This is required for the flow to work: the BankID app closes, and the user returns to the application in the background. Because of this behavior, having Active Login attach a nonce to the returnUrl isn’t straightforward.

That said, I agree this is an area worth revisiting, and we should explore potential improvements in the future.

PR #500, opened 2025-03-27
Feature/479 nonce

Zonnex:
This PR adds nonce and nonce-validation.

Created as draft for review. Piggy-backing on state-storage feature because we use IMemoryCache here as well.


elinohlsson:
In many scenarios (depending on device, OS, etc.), Active Login deliberately clears the returnUrl value and replaces it with an empty string. This is required to make the flow work, the BankID app closes, and the user returns to the application in the background. Because of this behavior, letting Active Login attach a nonce to the returnUrl isn’t straightforward. I think we need to handle this another way. Will close this PR for now and open a new issue for potential improvements in this area.

PR #517, opened 2025-10-08
fix(#479): store uiOptions in cookie

### LINKED ISSUE #516: Auth errors after upgrading to v11.1.0 - "Invalid returnUrl"
**Describe the bug**
After upgrading to v11.1.0 the auth flow sometimes fails with the error message "BankID auth failed with the error 'InvalidParameters' and the details 'Invalid returnUrl'"

UI

**Screenshots**

Image

**NuGet package version**
v11.1.0

**Runtime version**
.NET Core 8

Zonnex:
This PR fixes #516.

As per the issue linked, sometimes we send returnUrls to BankId that are invalid, one case seem to be that the returnUrl is too long which in our internal testing resulted in cryptic error, but others have received "Invalid Parameter". We couldn't replicate this directly but it's definitely a problem if the uiOptions are too long, which is also ugly.

Later we have a solution for those who wish to store data somewhere else but to fix issue we went with this approach for now, that we store the UIOptions in a cookie and pass a guid in the returnUrl instead. I have not had time to test this thoroughly so that's remaining.


Zonnex:
Yeah well that makes sense. Time to fix the tests

PR #526, opened 2025-11-11
Adhere to more modern BankID recommendations for opening in mobile device

### LINKED ISSUE #533: Unable to authenticate with Edge + Intune security controlled devices
**Describe the bug**
When authenticating with sameDevice using Edge, using Microsoft Intune controlled phones(iOS).
1) It asks "This site is trying to  open another application." Click Allow
2) Popup: "Something went wrong. The application couldn't be opened".
3) Then it stays spinning on the login screen, where "Avbryt" is your only option -> this leads to the "No such order" error sort of described in #508 , because the flow has not been properly initiated, there is no order. Switching manually to BankID app does not help. It has not been triggered.

We do not control the Intune settings so I cannot say which is causing it, but some security restriction must be.

However - attempting locally to making the library open according to BankIDs documentation (https links instead of bankid:// links - it works without issues).

UI

**To Reproduce**
1. Go to demo.activelogin.net with an affected device
2. Click on Login
3. Click on Use BankID  on this device
4. It asks "This site is trying to  open another application." Click Allow
5. Error : "Something went wrong. The application couldn't be opened".

**Expected behavior**
BankID being opened.

**NuGet package version**
What version of the Nuget package are you using? Latest and demo-site is uncoupled from what we use.

**Runtime version**
Are you using .NET Core / .NET Framework? What version? 
.NET 8, and whatever Demo siteuses:

 - Device: iPhones (several different tested)
 - OS: iOS 26 but also older
 - Browser : Edge
 - Version: 143.0.3650.112

urizen-source:
PR fixes #533

We have been able to reproduce issue with Edge+Intune that are security restricted, this fixes the issue. It also emulates the behaviour we see on test.bankid.com and follows the BankID documentation. If I understand correctly, the No such orders is when it thinks it has created a request to BankID(using bankid://) but it has been blocked, so a cancel or other action without a real order gives the "no such order" message.

Stops use of bankid:// on mobile devices  that are often security restricted. Instead uses the https://app.bankid.com,
Source: https://developers.bankid.com/getting-started/autostart 
Quote: "To open the Mobile BankID app from a browser on a mobile device, you should use unversal links for iOS and app links for Android. "

Displays fallback button by default on mobile devices:
Source: https://developers.bankid.com/resources/ui-guide-mobile
Quote: "4. Promt user to open BankID. The BankID app should be opened automatically using [autostart](https://developers.bankid.com/getting-started/autostart). **However, in case this fails for some reason, the user should be asked to open the app manually.**"

### DIFF
--- a/src/ActiveLogin.Authentication.BankId.AspNetCore/Client/activelogin-main.ts
+++ b/src/ActiveLogin.Authentication.BankId.AspNetCore/Client/activelogin-main.ts
@@ -109,6 +109,15 @@ function activeloginInit(configuration: IBankIdUiScriptConfiguration, initState:
     var flowIsCancelledByUser = false;
     var flowIsFinished = false;
 
+    function launchBankIdApp(url: string) {
+        // Use BankID recommended approach for launching the app
+        // See: https://developers.bankid.com/getting-started/autostart
+        const link = document.createElement("a");
+        link.href = url;
+        link.referrerPolicy = "origin";
+        link.click();
+    }
+
     function enableCancelButton(requestVerificationToken: string, cancelUrl: string, orderRef: string = null) {
         var onCancelButtonClick = (event: Event) => {
             cancel(requestVerificationToken, cancelUrl, orderRef);
@@ -132,7 +141,7 @@ function activeloginInit(configuration: IBankIdUiScriptConfiguration, initState:
 
                     if (data.deviceMightRequireUserInteractionToLaunchBankIdApp) {
                         var startBankIdAppButtonOnClick = (event: Event) => {
-                            window.location.href = data.redirectUri;
+                            launchBankIdApp(data.redirectUri);
                             hide(startBankIdAppButtonElement);
                             event.target.removeEventListener("click", startBankIdAppButtonOnClick);
                         };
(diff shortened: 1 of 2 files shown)


okalangkenneth:
Some supporting evidence for this approach.

Looking into #543 (Edge on Android failing to open the BankID app) and #535 (Opera on Android failing to complete authentication after launch), both appear to be symptoms of the same underlying gap.... the current Chrome-only gate on App Links leaves other Chromium-based browsers without a working launch path.

This PR's approach of switching all mobile devices to https://app.bankid.com aligns with BankID's own documentation and would resolve both of those issues without needing per-browser special-casing going forward.

All checks are passing, and the changes look like they can be cleanly merged after a rebase. I can help with that or add test coverage if that would move this forward.

elinohlsson:
Thanks a lot for your contribution and all the work on this @urizen-source!

We had to make some changes before we could release the solution, so we ended up creating a new PR inspired by your work here. Active Login now uses [https://app.bankid.com](https://app.bankid.com) for mobile devices.

However, we had to make an exception for Samsung Internet, Firefox, and Edge on Android. With these browsers, we were not able to get the complete launch and redirect flow working reliably using [https://app.bankid.com](https://app.bankid.com), so we kept the existing approach for these cases.

This is implemented and released in **v12.1.0**.

See the [release notes](https://github.com/ActiveLogin/ActiveLogin.Authentication/releases/tag/v12.1.0) for details.

I will now close this PR. If you still experience problems related to this after upgrading to **v12.1.0**, please open a **new issue** rather than commenting on this PR. Comments on closed PRs and issues may not be noticed or followed up.

When opening the new issue, you can of course include a reference to this PR so we can connect it to the original work.

PR #542, opened 2026-02-19
Made the ctor public.

### LINKED ISSUE #541: Public constructor for BankIdClaimsTransformationContext
When adding support for the Phone Auth API I want to ensure the same behavior as the standard authentication support, including the same claims transformation. The problem is that I can't create the BankIdClaimsTransformationContext since the constructor is internal.

AspNetCore

Make the constructor public.

The only alternative is to re-implement everything related to claims transformation.

equist:
This PR fixes #541.

- [x] The constructor of BankIdClaimsTransformationContext is made public


torselden:
Nice and low-risk change. Making the constructor public is a sensible, backwards-compatible improvement for custom claims transformation scenarios, and the change is narrow and easy to reason about. The ASP.NET Core test suite still passes, so this looks mergeable as-is. I support merging this.

PR #547, opened 2026-08-19
Respect RouteOptions.LowercaseUrls for BankID UI redirects

### LINKED ISSUE #523: RouteOptions.LowercaseUrls isn't respected
**Describe the bug**

When setting `RouteOptions.LowercaseUrls` to `true` this package still uses camel casing.

API

**To Reproduce**

Set `LowercaseUrls` to `true`:

```csharp
builder.Services.Configure<RouteOptions>(options =>
{
    options.LowercaseUrls = true;
});
```

After challenge when `/ActiveLogin/BankId/Auth?returnUrl=%2Fal-bankid-auth-samedevice-callback&uiOptions=...` is called, it is CamelCase instead of lowercase.

**Expected behavior**
Request `/activelogin/bankid/auth`

**Screenshots**
N/A

**NuGet package version**
11.0.0

**Runtime version**
.NET 9

 - Device: iPhone 8 and 16 Pro
 - OS: 16.7.10 and 18.6.2
 - Browser: Safari, works on Chrome

**Additional context**
We use Razor Pages and not MVC. We noticed this when trying to figure out why BankID was launched two time on Safari on iOS. It will get called again but the next time it's lowercase. Not sure yet if it's related.

torselden:
The auth challenge redirect and the payment/sign init redirects built their URLs by hand, always producing camel-cased paths and ignoring RouteOptions.LowercaseUrls. The BankID app return URL is generated via routing (Url.Action) and does respect it, so with LowercaseUrls enabled the two URLs differed only by case. On Safari/iOS this caused the BankID app to launch twice.

This PR fixes #523 

- [X] Generate the init URLs via LinkGenerator so they honor routing conventions and stay consistent with the return URL.

- [X] Code written
- [X] Test added
- [ ] Documentation updated / written **(not needed in this case)**


rfolkes:
Looks good! - we should add  .vscode/mcp.json to gitignore as well - can you add that as well in this PR @torselden

torselden:
Thanks! I've added mcp.json to the .gitgnore

PR #548, opened 2026-08-20
Align BankID app autostart with current BankID guidelines

### LINKED ISSUE #534: Review how Active Login performs autostart of the BankID security application
### Description

Review how Active Login performs autostart of the BankID security application, including which links are used, to conform with the current recommendations from BankID:
https://developers.bankid.com/how-to-guides/autostart

### Context

BankID has updated its guidance on how the autostart link should be constructed and used when launching the BankID app since this functionality was implemented in Active Login. The current implementation may not fully align with these recommendations.

### Related issues and PRs

- Related issue: #533
- Related PR: #526

torselden:
Updates how Active Login launches the BankID app so it follows BankID's current [autostart guidance](https://developers.bankid.com/how-to-guides/autostart). Autostart is the default on all supported devices, the recommended universal/app link is used on mobile, and the manual launch button is only shown as a fallback where a browser actually requires user interaction.

This PR fixes #534 

**High level overview of this PR**
- Use the universal/app link (https://app.bankid.com/) for all iOS and Android browsers, and keep the bankid:/// scheme for desktop
- Launch the app via BankID's recommended anchor-click with referrerPolicy="origin" instead of window.location.href (fixes unreliable iOS universal-link launch)
- Restrict the manual "start BankID" fallback button to Android browsers that require a user gesture (Chrome, Edge, Samsung Internet, Brave); keep autostart automatic on iOS and on Android Firefox/Opera

These things have been implemented (when relevant)
- [x] Code written
- [X] Test added — launcher tests for app-link vs scheme selection, the user-interaction matrix, and redirect parameter behavior (Android redirect=null, iOS Safari return URL, iOS Chrome/Firefox browser scheme, iOS Edge/Opera redirect=null)
- [X] Documentation updated / written — added a "How the BankID app is launched" section and a mobile browser-support note in bankid.md

**Backwards-compatibility note**
This PR changes the launch URL selection for several device/browser combinations (no public API changes):

- CanUseAppLink no longer restricts the app link to Safari-on-iOS and Chrome/Edge-on-Android ≥ 6 — it now applies to every iOS and Android browser. The DeviceOsVersion.MajorVersion >= 6 guard was removed.
As a result, iOS non-Safari (Chrome, Firefox, Edge, Opera) and Android Firefox/Opera / older Android now receive https://app.bankid.com/… where they previously received bankid:///…. This matches current BankID guidance and should improve reliability.

- The manual "start BankID" fallback button is now scoped to restricted Android browsers instead of a broader set.
What consumers should verify: if you relied on the old scheme-based launch for a specific browser, that path changes. Android Firefox/Opera in particular should be verified on real devices, since their historical app-link/intent handling has been inconsistent. Custom IBankIdLauncher / IBankIdSupportedDeviceDetector implementations are unaffected, and the deprecated redirect query parameter is still emitted, so BankID app compatibility is preserved.

### DIFF
--- a/docs/articles/bankid.md
+++ b/docs/articles/bankid.md
@@ -1394,6 +1394,19 @@ services.AddTransient<IBankIdQrCodeGenerator, CustomQrCodeGenerator>();
 
 The functionality provided tries to detect the device by looking at the user agent. We need to know what device is used to launch the BankId app and this differs from iOS/Android/PC/Mac.
 
+#### How the BankID app is launched
+
+By default, Active Login follows [BankID's autostart guidance](https://developers.bankid.com/how-to-guides/autostart):
+
+* **Mobile (iOS and Android):** the app is launched using the universal/app link `https://app.bankid.com/`.
+* **Desktop:** the app is launched using the custom scheme `bankid:///`.
+
+The launch link is opened from the client using an anchor element with `referrerPolicy="origin"`, as recommended by BankID.
+
+Autostart is the default. A manual *"Start the BankID app"* button is only shown as a fallback on Android browsers that block launching a third party app without a user gesture (Chrome, Edge, Samsung Internet and Brave). Firefox and Opera on Android, as well as all browsers on iOS, autostart without user interaction.
+
+The deprecated `redirect` parameter is still included in the launch URL for backwards compatibility. BankID now recommends providing the return URL in the backend call when creating the order instead, see [Return URL](https://developers.bankid.com/how-to-guides/return-url).
+
 By implementing `IBankIdLauncher` you can customize exactly how to launch the app. It is very rare that you need to change this, but could be relevant if you use Active Login for authenticating a user in a native mobile app.
 
 ```csharp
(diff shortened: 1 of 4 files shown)


elinohlsson (approved):
Looks really good! I will just do some manual testing before merging.

elinohlsson:
I made a couple of changes based on testing the BankID autostart flow across our supported mobile browsers:

- Keep BankID's recommended App Link approach (`https://app.bankid.com/`) as the default for mobile.
- On Android, Firefox and Samsung Internet are explicit exceptions. Our testing shows that the App Link does not launch the BankID app in these browsers, so they fall back to the `bankid:///` scheme.
- I tested reverting the new JavaScript launch implementation to `window.location.href`, but that did not change the behavior. This suggests the issue is specific to App Link support in these browsers rather than `launchBankIdApp()`.
- Updated the browser support documentation to describe the Android exceptions.
- Removed the outdated IE11/polyfill documentation.
- Removed Opera from the verified browser list for now, since the redirect flow is currently not working there.

With these changes, the BankID launch flow works for all of the Android browsers we currently test before release, as well as all of the iOS browsers we test.

There are still some existing redirect issues that need to be addressed, particularly with Edge, and Opera support needs to be investigated. These issues were present before this PR and are not introduced by these changes, so I suggest we look into them separately in another PR.

PR #553, opened 2026-08-20
Add configurable collect states to the simulated BankID environment

### LINKED ISSUE #397: Add possibility to provide CollectStates in UseSimulatedEnvironment
The time it takes to authenticate using UseSimulatedEnvironment is over 10 seconds. In one way it is a good way to illustrate how long it takes in real life and you get to see some different CollectStates in the UI. However, when you use the simulated environment in test environments or unit/integration tests it gets cumbersome to have to wait that long.

Core

One possible solution would be to add a parameter to UseSimulatedEnvironment to be able to set the List<CollectState> collectStates that is exposed in one of the constructors of BankIdSimulatedApiClient.

To make it easier some defaults could be exposed as a static, for example 

* Normal - tries to imitate a "normal" flow
* All - Includes all the collect states to be able to verify that the ui can handle all different text lengths
* Fast - As few states as possible to make the sign in as fast as possible.

Today it is possible to do this by implementing your own verison of UseSimulatedEnvironment by copying the private method from here, https://github.com/ActiveLogin/ActiveLogin.Authentication/blob/main/src/ActiveLogin.Authentication.BankId.Core/IBankIdBuilderExtensions.cs#L217-L234, and pass in your own instance of BankIdSimulatedApiClient

torselden:
This PR fixes #397.

- [x] Allow `UseSimulatedEnvironment` to configure the collect-state sequence used during simulated authentication.
- [x] Add reusable `Normal`, `Fast`, and `AllSuccessful` collect-state sequences while preserving the existing default behavior.
- [x] Make custom collect-state configuration safer by validating the input and avoiding unexpected changes when the caller mutates the original list.
- [x] Clarify that `AllSuccessfulCollectStates` covers successful UI states and intentionally excludes failed terminal states.
- [x] Keep response-delay configuration outside the scope of this issue.

These things have been implemented:

- [x] Code written
- [x] Test added

### DIFF
--- a/docs/articles/bankid.md
+++ b/docs/articles/bankid.md
@@ -225,7 +225,7 @@ services
     });
 ```
 
-The `AllCollectStates` sequence can be used to verify that the UI handles the different successful collect hint codes and text lengths.
+The `AllSuccessfulCollectStates` sequence can be used to verify that the UI handles the different successful collect hint codes and text lengths. It intentionally does not include failed terminal states, so the simulated authentication can complete successfully.
 
 ```csharp
 services
@@ -233,7 +233,7 @@ services
     {
         bankId.UseSimulatedEnvironment(options =>
         {
-            options.CollectStates = BankIdSimulatedAppApiClient.AllCollectStates;
+            options.CollectStates = BankIdSimulatedAppApiClient.AllSuccessfulCollectStates;
         });
     });
 ```
@@ -255,7 +255,7 @@ services
     });
 ```
 
-These examples require the `ActiveLogin.Authentication.BankId.Api` and `ActiveLogin.Authentication.BankId.Core` namespaces. The simulated client also applies a response delay to mimic the real flow. Set `Delay` on the resolved `BankIdSimulatedAppApiClient` when a test needs to remove that delay completely.
+These examples require the `ActiveLogin.Authentication.BankId.Api` and `ActiveLogin.Authentication.BankId.Core` namespaces. The simulated client also applies a response delay to mimic the real flow. Delay is not configured by `UseSimulatedEnvironment`; if a test needs to remove it, resolve the registered `BankIdSimulatedAppApiClient` from dependency injection and set its `Delay` property to `TimeSpan.Zero`.
 
 
 ### Test environment
--- a/docs/articles/bankid.md
+++ b/docs/articles/bankid.md
@@ -210,6 +210,54 @@ services
 ```
 
 
+### Simulated environment with custom collect states
+
+The simulated environment uses a normal collect flow by default. For automated tests, a faster flow can be selected. The `FastCollectStates` sequence contains only the states needed to reach a successful result.
+
+```csharp
+services
+    .AddBankId(bankId =>
+    {
+        bankId.UseSimulatedEnvironment(options =>
+        {
+            options.CollectStates = BankIdSimulatedAppApiClient.FastCollectStates;
+        });
+    });
+```
+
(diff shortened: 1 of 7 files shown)


docs/articles/bankid.md
```diff
@@ -210,6 +210,54 @@ services
...
+    .AddBankId(bankId =>
+    {
+        bankId.UseSimulatedEnvironment(options =>
+        {
+            options.CollectStates = BankIdSimulatedAppApiClient.FastCollectStates;
+        });
+    });
+```
+
+The `AllSuccessfulCollectStates` sequence can be used to verify that the UI handles the different successful collect hint codes and text lengths. It intentionally does not include failed terminal states, so the simulated authentication can complete successfully.
```
Abrissirba: Feels a bit much to give a fulle code example just to show another option.

src/ActiveLogin.Authentication.BankId.Core/IBankIdBuilderExtensions.cs
```diff
@@ -267,7 +267,35 @@ public static IBankIdBuilder UseProductionEnvironment(this IBankIdBuilder builde
...
+        }
+
+        var options = new BankIdSimulatedEnvironmentOptions();
+        configure(options);
+        if (options.CollectStates == null || options.CollectStates.Count == 0)
+        {
+            throw new ArgumentException("At least one collect state must be configured.", nameof(configure));
+        }
+
+        var collectStates = new List<BankIdSimulatedAppApiClient.CollectState>(options.CollectStates);
```
Abrissirba: Why do we copy the collect states both here and in the BankIdSimulatedAppApiClient?

src/ActiveLogin.Authentication.BankId.Core/IBankIdBuilderExtensions.cs
```diff
@@ -267,7 +267,35 @@ public static IBankIdBuilder UseProductionEnvironment(this IBankIdBuilder builde
...
+    public static IBankIdBuilder UseSimulatedEnvironment(this IBankIdBuilder builder, Action<BankIdSimulatedEnvironmentOptions> configure)
+    {
+        if (configure == null)
+        {
+            throw new ArgumentNullException(nameof(configure));
+        }
+
+        var options = new BankIdSimulatedEnvironmentOptions();
+        configure(options);
+        if (options.CollectStates == null || options.CollectStates.Count == 0)
```
Abrissirba: Do we need this validation here when we have it in the BankIdSimulatedAppApiClient constructor?

src/ActiveLogin.Authentication.BankId.Core/IBankIdBuilderExtensions.cs
torselden: We copy the collect-state list to avoid shared mutable state. The configuration object and the simulated client both hold a list, and we want each layer to own its own snapshot. Otherwise, a caller can mutate the original list after construction and unexpectedly change the behavior of the simulated BankID flow. To defensive perhaps?

Abrissirba: A bit too defensive for my taste but it is fine :) One alterantive might be to take a IReadOnlyList as parameter instead.

torselden: Removed builder-level defensive copy, kept client-level snapshot copy because that is the actual ownership boundary.

